PR #170, opened 2014-03-07
Split agent reply timeout from execution timeout

ncdc:
If the agent's action times out in the middle of sending a reply, it may
send a partial message over the wire, which can lead to errors in the
messaging broker.

Run the agent's action in its own timeout block, and if it finishes,
yield the replies in a separate timeout block. This way, it should be
less likely that a partial message is sent out over the wire.

### DIFF
--- a/lib/mcollective/agents.rb
+++ b/lib/mcollective/agents.rb
@@ -122,9 +122,12 @@ module MCollective
         begin
           agent = PluginManager["#{request.agent}_agent"]
 
+          replies = nil
           Timeout::timeout(agent.timeout) do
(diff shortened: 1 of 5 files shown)


MikaelSmith:
@ncdc did you end up reviewing the CLA? I'm not sure if this change was ever made, and @geekygirldawn isn't around to help me follow up.

Since this adds a new configuration option, I'd like to see an MCO ticket for it. Also, you'll need a rebase.

ncdc:
I'm not actively working on MCO any more. It's ok to close.

PR #270, opened 2014-11-26
DO NOT MERGE - MCO-517 AIO paths

richardc:
This PR is here so we can have travis coverage, and can easily find the code/collaborate.

Don't merge (yet)

### DIFF
--- a/bin/mcollectived
+++ b/bin/mcollectived
@@ -20,7 +20,7 @@ daemonize = nil
 if MCollective::Util.windows?
   configfile = File.join(MCollective::Util.windows_prefix, "etc", "server.cfg")
 else
-  configfile = "/etc/mcollective/server.cfg"
+  configfile = "/etc/puppetlabs/agent/mcollective/server.cfg"
 end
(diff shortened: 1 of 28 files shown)


haus:
@richardc with the cfacter and facter work, @whopper made it so that we would try the new location first, and fall back to the old location. that seems like it would be a good idea here as well.

richardc:
@haus New and old location of what?

haus:
@richardc sorry, wasn't specific. Things like configfile, where we could check file existence of the new and fall back to the old. Specifically, I was thinking of `bin/mcollectived` and `lib/mcollective/util.rb` as the places where that makes sense in the current PR.

richardc:
I'd sooner we just fail to load a config file if it isn't in the new location, rather than play a round of 'guess which config file will be respected'

richardc:
A new branch for this has been created as we have multiple contributors/contributions to manage - https://github.com/puppetlabs/marionette-collective/tree/aio-agent

PR #277, opened 2015-01-22
(MCO-550) Use RubyW for service

Iristyle:
Previously, any powershell related functions (facter facts, providers, execs,
MCO triggered powershell commands) fail to return 0 properly, and therefore
fail, but only when triggered via the MCO daemon under windows.

Switching back to using rubyw now that bug #7239 was backported to Ruby
1.9.3 (https://bugs.ruby-lang.org/issues/9232). This allows us to take
advantage of using rubyw and providing proper exit codes for powershell
processes.

### DIFF
--- a/ext/windows/daemon.bat
+++ b/ext/windows/daemon.bat
@@ -3,4 +3,4 @@ SETLOCAL
 
 call "%~dp0..\bin\environment.bat" %0 %*
 
-ruby -rubygems "%MCOLLECTIVE_DIR%\bin\mcollectived" %*
+rubyw -rubygems "%MCOLLECTIVE_DIR%\bin\mcollectived" %*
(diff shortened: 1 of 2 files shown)


richardc:
Please retarget this to master.

ploubser:
@Iristyle We'll be cutting a 2.8 for AIO so the previous PR targeted at master will be going in. Closing for now.

Iristyle:
Ok, didn't know if there was going to be anything else in the 2.7.z line.  There was some confusion over whether we needed to backport the fix.

If there is going to be a 2.7.1 release, I would definitely lobby that this fix goes in as it's high impact.  If there are no plans for another release in the 2.7 line, then this can be ignored.

ferventcoder:
:+1: that this gets into next release, whether 2.8 or 2.7.z :)

joshcooper:
Superceded by PR#271

kylog:
No plans for another foss 2.7.x release.

kylog:
Hah. I thought this was recent. Never mind me chiming in a year later :(

ferventcoder:
Ha!

PR #282, opened 2015-01-29
Allow mco inventory to work with servers <2.1.0 that don't return any data plugins

craigmiskell:
Otherwise we get:
   Data Plugins:
Failed to display node inventory: NoMethodError: undefined method `size' for nil:NilClass


richardc:
Thanks for the contribution but I'm going to drop it; there are many other ways that 2.7 (soon to be 2.8) aren't compatible with older mcollective installs, so this would be the start of trying to paper over the grand canyon.

PR #285, opened 2015-01-29
MCO-555 Add AIO variants of init and sysconfig files

richardc:
The All-In-One Agent configures MCollective with a specific set of
paths for configurations and pidfiles.

Here we take the work of Geoff Nichols from PR#274 and moves it
into ext/aio, where the All-In-One build process can consume the
files.

### DIFF
--- a/ext/aio/debian/mcollective.init
+++ b/ext/aio/debian/mcollective.init
@@ -1,0 +1,92 @@
+#!/bin/sh
+#
+# mcollective   Application Server for STOMP based agents
+#
+#
+# description:  mcollective lets you build powerful Stomp compatible middleware clients in ruby without having to worry too
+#               much about all the setup and management of a Stomp connection, it also provides stats, logging and so forth
+#               as a bonus.
+#
+### BEGIN INIT INFO
+# Provides:          mcollective
+# Required-Start:    $remote_fs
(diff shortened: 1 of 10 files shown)


ext/aio/debian/mcollective.init
```diff
@@ -0,0 +1,92 @@
...
+
+uid=`id -u`
+[ $uid -gt 0 ] && { echo "You need to be root to run file" ; exit 4 ; }
+
+
+# PID directory
+pidfile="/var/run/puppetlabs/agent/mcollectived.pid"
+
+name="mcollective"
+mcollectived=/opt/puppetlabs/agent/bin/mcollectived
```
geoffnichols: @richardc Will there be an update to install.rb to use bin (instead of sbin) for mcollectived? cc/ @haus

richardc: AIO uses install.rb? Then yes I'll add an option to install.rb to use bin for mcollectived.

richardc: Looking at install.rb no modifications are required.

When invoking install.rb add `--sbindir=/opt/puppetlabs/agent/bin`

geoffnichols:
:+1:

PR #293, opened 2015-02-09
(MCO-591) Fix condrestart on debian for AIO agent

branan:
condrestart was built around some sort of busted lock file
functionality. This changes it to rely on start-stop-daemon to
determine if the process is running or not.


ext/aio/debian/mcollective.init
```diff
@@ -72,11 +71,11 @@ case "$1" in
...
     condrestart)
-        if [ -f $lock ]; then
-            $0 stop
-            # avoid race
-            sleep 2
-            $0 start
+	# The debian status helper just aborts the script if the
+	# process is not running, so we call start-stop-daemon
+	# explicitly
+        if start-stop-daemon -T --oknodo -q -p ${pidfile}; then
```
haus: can we use the long flags here for readability?

haus: `--status`, `--quiet`, `--pidfile` i imagine

haus:
:+1:

PR #296, opened 2015-02-16
MCO-594 Update for latest AIO paths

richardc:
The latest version of the Unified Puppet Agent Layout calls for config files in
/etc/puppetlabs/$project/$project.conf, logfiles in
/var/log/puppetlabs/$project.log, pidfiles in /var/run/puppetlabs/$project.pid,
binaries in /opt/puppetlabs/puppet/bin and user plugins for MCollective in
/opt/puppetlabs/mcollective/plugins.

Here we make those changes.

### DIFF
--- a/ext/aio/debian/mcollective.init
+++ b/ext/aio/debian/mcollective.init
@@ -24,11 +24,11 @@ uid=`id -u`
 
 
 # PID directory
-pidfile="/var/run/puppetlabs/agent/mcollectived.pid"
+pidfile="/var/run/puppetlabs/mcollectived.pid"
 
 name="mcollective"
-mcollectived=/opt/puppetlabs/agent/bin/mcollectived
-daemonopts="--pid=${pidfile} --config=/etc/puppetlabs/agent/mcollective/server.cfg"
+mcollectived=/opt/puppetlabs/puppet/bin/mcollectived
+daemonopts="--pid=${pidfile} --config=/etc/puppetlabs/mcollective/server.cfg"
 
 
 # Source function library.
--- a/bin/mcollectived
+++ b/bin/mcollectived
@@ -22,7 +22,7 @@ if MCollective::Util.windows?
   configfile = File.join(MCollective::Util.windows_prefix, "etc", "server.cfg")
 else
   # search for the server.cfg in the AIO path then the traditional one
-  configfiles = ['/etc/puppetlabs/agent/mcollective/server.cfg',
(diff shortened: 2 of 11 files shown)


ext/aio/debian/mcollective.init
```diff
@@ -24,11 +24,11 @@ uid=`id -u`
 
 
 # PID directory
-pidfile="/var/run/puppetlabs/agent/mcollectived.pid"
+pidfile="/var/run/puppetlabs/mcollectived.pid"
 
 name="mcollective"
-mcollectived=/opt/puppetlabs/agent/bin/mcollectived
-daemonopts="--pid=${pidfile} --config=/etc/puppetlabs/agent/mcollective/server.cfg"
+mcollectived=/opt/puppetlabs/puppet/bin/mcollectived
```
richardc: The AIO guide calls for

* Binaries live in /opt/puppetlabs/puppet/bin

Which is what's been done here, but it's a little weird to have `/opt/puppetlabs/puppet/bin/mcollectived` and `/opt/puppetlabs/puppet/bin/mco`.

kylog: Good point.

It's worth asking whether we could have single (rather than per-project) directories for bin/lib/include. I.e. `/opt/puppetlabs/bin`, etc. Seems like we'd have path pollution if we had a number of per-project directories, and if we had collisions in binary names or library names, etc, that we'd be asking for trouble anyway.

But also, I'm not sure what PE paths for bin/lib/include are like.

/cc @haus @stahnma

joshcooper: I believe `/opt/puppetlabs/bin` is correct as it will have all puppet-related executables including `facter`, `cfacter`, `ruby`, `openssl`.

haus: /opt/puppetlabs/bin will have symlinks of all of our projects, such as puppet, facter, cfacter, hiera, mco. It will not have links of ruby or openssl. It is intended to be relatively safe to add to PATH.

/cc @ahpook  @shrug

richardc: As /opt/puppetlabs/bin doesn't have ruby or openssl it may also make sense to exclude symlinks for mcollectived from this path, as it's not expected that a user will run mcollectived as an interactive command.

This in addition to PUP-3809 has answered my question about the correctness of /opt/puppetlabs/puppet/bin/mcollectived, so I'm happy with this line and seek further review.

Thanks

kylog:
+1 to all that. Makes sense.

PR #300, opened 2015-03-05
MCO-617 Assign version in-tree

richardc:
Here we change version assignment from a release machinery task to a manual one, for consistency with other PuppetLabs projects.

### DIFF
--- a/ext/project_data.yaml
+++ b/ext/project_data.yaml
@@ -6,7 +6,6 @@ homepage: 'https://docs.puppetlabs.com/mcollective/'
 summary: 'Application Server for hosting Ruby code on any capable middleware'
 description: 'The Marionette Collective, e.g. mcollective, is a framework for building server orchestration or parallel job execution systems.'
 version_file: 'lib/mcollective.rb'
-update_version_file: true
 # files and gem_files are space separated lists
 files: 'mcollective.init COPYING doc etc lib plugins ext bin install.rb'
 # List of packaging related templates to evaluate before the tarball is packed
--- a/lib/mcollective.rb
+++ b/lib/mcollective.rb
@@ -59,7 +59,7 @@ module MCollective
 
   MCollective::Vendor.load_vendored
 
-  VERSION="@DEVELOPMENT_VERSION@"
+  VERSION="2.8.0"
 
   def self.version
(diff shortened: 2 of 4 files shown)


Iristyle:
FWIW - this looks good to me.  The only other thing I would say is that the rake tasks used to generate JIRA tix for an `MCO release` should be updated so that RelEng knows to bump the  file instead of simply adding an annotated git tag.

/cc @kylog @haus

richardc:
Apart from I still want there to be an annotated tag so I can continue to do things like this: https://github.com/puppetlabs/marionette-collective/compare/2.7.0...2.8.0

Iristyle:
@richardc agreed - I should have been a little more clear.  We still wanted the annotated tag for sure.

richardc:
Releng doesn't bump the file, that's $DEVELOPER, https://github.com/puppetlabs/packaging/blob/master/tasks/tickets.rake#L292-L294  and I'm fine with that.

kylog:
Agreed. I was thinking that jira task (done by dev) would cover the version bump.

:+1:

richardc:
Which is to say, we already have that task getting created by the release-ticket maker, it's just in previous MCollective releases I've been closing it with 'the tooling does this'

Iristyle:
:+1: just wanted to make sure nothing slips through the cracks during release - but looks like we're good already!

ext/project_data.yaml
```diff
@@ -6,7 +6,6 @@ homepage: 'https://docs.puppetlabs.com/mcollective/'
 summary: 'Application Server for hosting Ruby code on any capable middleware'
 description: 'The Marionette Collective, e.g. mcollective, is a framework for building server orchestration or parallel job execution systems.'
 version_file: 'lib/mcollective.rb'
```
haus: You can also lose the version_file if you want, but leaving it won't do any harm.

kylog:
@richardc merged. But I don't know your procedure for merge-ups to master/etc, so I did nothing as far as merge-up.

ext/project_data.yaml
Iristyle: :+1: to remove if it's just a vestigiality...

PR #305, opened 2015-03-23
(maint) pin systemu on ruby 1.8

richardc:
As per https://github.com/ahoward/systemu/issues/41 systemu's ruby
support policy is "whatever Ruby version is officially supported
at the point in time of the release"

For 1.8 pin to systemu 2.6.4, which is the last known working version
for 1.8

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,7 +2,12 @@ source 'https://rubygems.org'
 
 gem 'json'
 gem 'stomp'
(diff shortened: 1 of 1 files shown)


richardc:
Should fix the travis failures as observed in #304's builds:  https://travis-ci.org/puppetlabs/marionette-collective/jobs/55346883

PR #306, opened 2015-04-09
[DO NOT MERGE](MCO-625) mco.bat should use ruby instead of %RUBY%

ferventcoder:
The `mco.bat` file built with the MSI uses `%RUBY%` rather than `ruby`
used by `puppet.bat` or `facter.bat`. This environmental variable
does not appear to be set in `environment.bat`.

This causes calls to `mco` on Windows to return the following error:

```
C:\Users\Administrator>"C:\Program Files\Puppet Labs\Puppet\bin\mco.bat" ping
'-S' is not recognized as an internal or external command, operable program
or batch file.
```

Other bat files used by the puppet agent do not display this similar
behavior. Fix up mco.bat to use ruby instead of %RUBY%. Without this
change, calls to `mco` on Windows will continue to fail.

### DIFF
--- a/ext/windows/mco.bat
+++ b/ext/windows/mco.bat
@@ -4,4 +4,4 @@ SETLOCAL
 
 call "%~dp0environment.bat" %0 %*
 
-%RUBY% -S -- mco %* --config "%CLIENT_CONFIG%"
+ruby -S -- mco %* --config "%CLIENT_CONFIG%"


ferventcoder:
I'm gonna guess the Travis CI failure is related to a pre-existing issue with 1.8.7 and stable?

ferventcoder:
This has also been created https://github.com/puppetlabs/puppet_for_the_win/pull/107

richardc:
Yes, fixed on master with: https://github.com/puppetlabs/marionette-collective/pull/305

Why is this targetting stable?

ferventcoder:
Targeting stable because I didn't know what to target, but would need to go out with next version.

ferventcoder:
Closing to make more enhancements to PFW after discussion.

PR #307, opened 2015-04-13
(maint) Remove Fedora 19 from build targets

melissa:
Fedora 19 went EOL on 2015-01-06. We should not longer be providing
packages for this platform.


melissa:
@ploubser and @richardc I wasn't positive where to point this PR. I'll open up similar ones against master/stable. Feel free to close any that aren't needed.

PR #311, opened 2015-04-24
FOR DISCUSSION (maint) include the 'classic' libdirs in the default config

richardc:


richardc:
This will need a better commit message, will re-work and re-open.

PR #312, opened 2015-05-04
(MCO-646) Add launchd plist for use with OSX AIO

heathseals:
This commit adds a launchd plist for use with OSX AIO packaging.


heathseals:
Removed RUBYLIB. It doesn't appear to be required by AIO ruby.

melissa:
:+1:

PR #318, opened 2015-06-02
(MCO-671) Add acceptance tests to mco

johnduarte:
This commit adds the infrastructure to support running acceptance tests
against mcollective.


acceptance/setup/aio/pre-suite/070_Install_puppet-agent-plugin.rb
```diff
@@ -0,0 +1,26 @@
...
+   mco_libdir = '/opt/puppetlabs/mcollective/plugins/mcollective'
+   git_pkg = 'git'
+   if agent.platform =~ /ubuntu-10/
+     git_pkg = 'git-core'
+   end
+   install_package(agent, git_pkg)
+  end
+  on agent, "mkdir -p #{mco_libdir}"
+  on agent, "git clone https://github.com/puppetlabs/mcollective-puppet-agent.git"
+  on agent, "cd mcollective-puppet-agent && for i in agent aggregate application data ext util validator ; do cp -a $i #{mco_libdir} ; done"
```
richardc: Don't need to copy ext.  That's just packaging metadata.

acceptance/files/ca_crt.pem
```diff
@@ -0,0 +1,32 @@
+-----BEGIN CERTIFICATE-----
```
richardc: Not specific to this file, but can we add a README.md to the acceptance/files directory using very strong language that the files here should not be used in production as they are by definition not secret secrets.

Also maybe worth adding the section from https://github.com/richardc/mcollective-aio-beaker#ssl-setup which describes how the files have been derived, in case they need to be regenerated in future.

acceptance/README.md
```diff
@@ -0,0 +1,26 @@
+beaker tests for validation of MCollective in puppet-agent
+
```
richardc: Would like some dire warnings that these files aren't secret secrets.

And a rad picture of a dragon.

The warning is fine.

johnduarte: @richardc sorry, I forgot to add the warning before. Please take a look at the updated README.md and see if is sufficient.

richardc: Sweet dragon

PR #322, opened 2015-06-24
(MCOP-521) mco runonce with powershell provider test

johnduarte:
This commit adds an acceptance test to verify that executing
a `mco puppet runonce` where a Windows agent is classified
with an exec resource that uses the powershell provider correctly
executes on the Windows system.

This test is currently expected to fail until MCOP-521 is resolved.
The test assertion is wrapped in an `expect_failure` method to
prevent the test run from turning red until MCOP-521 has been fixed.
At that time, the test will fail and the wrapper can be removed to
flip the result.

### DIFF
--- a/acceptance/tests/mco_puppet_powershell.rb
+++ b/acceptance/tests/mco_puppet_powershell.rb
@@ -1,0 +1,119 @@
+test_name "MCOP-521 mco puppet run with powershell provider" do
+
+  testdir = master.tmpdir('mco_powershell')
+  testfile = master.tmpfile('mco_powershell').split('/')[-1]
+
+  # gather array of windows hosts
+  windows_hosts = []
+  hosts.each do |h|
+    if /windows/ =~ h[:platform]
+       windows_hosts << h
+    end
+  end
+  if windows_hosts.empty?
+    skip_test "No windows hosts to test powershell on"
+  end
+
+
+  step "Add powershell manifest to windows host classification" do
+    node_str = ''
+    windows_hosts.each do |h|
+      n =<<-EOS
+        node #{h} {
+            exec { "create-test-file":
+              command => "out-file C:\\#{testfile}.txt",
+              provider => powershell,
+            }
+        }
+EOS
+      node_str << n
+    end
+
+    apply_manifest_on(master, <<-MANIFEST, :catch_failures => true)
+      File {
+        ensure => directory,
+        mode => "0750",
+        owner => #{master.puppet['user']},
+        group => #{master.puppet['group']},
+      }
+      file {
+        '#{testdir}':;
+        '#{testdir}/environments':;
+        '#{testdir}/environments/production':;
+        '#{testdir}/environments/production/manifests':;
+        '#{testdir}/environments/production/manifests/site.pp':
+          ensure => file,
+          mode => "0640",
+          content => '
+            #{node_str}
+            node default {}
+            ';
+      }
+MANIFEST
+
+  end
+
+  master_opts = {
+    'main' => {
+      'environmentpath' => "#{testdir}/environments",
+     }
+  }
+
+  with_puppet_running_on(master, master_opts) do
+    if mco_master.platform =~ /windows/ then
+      if mco_master[:ruby_arch] == 'x86' then
+        mco_bin = 'cmd.exe /c "C:\Program Files (x86)\Puppet Labs\Puppet\bin\mco.bat"'
+      else
+        mco_bin = 'cmd.exe /c "C:\Program Files\Puppet Labs\Puppet\bin\mco.bat"'
+      end
+    else
+     mco_bin = '/opt/puppetlabs/bin/mco'
+    end
+
+    step "Stub puppet and run agent to get certs"
+    hosts.each do |host|
+      stub_hosts_on(host, 'puppet' => master.ip)
+    end
+
+    step "Install powershell module on master"
+    install_puppet_module_via_pmt_on(master, {:module_name => "puppetlabs-powershell"})
+
+    step "Delete puppet cache dir on Windows hosts, because `agent` run created it with wrong permissions"
+      # NOTE: `with_puppet_running_on` triggers a `puppet agent -t` run on the
+      # Windows agents using the ADMINISTRATOR user. This creates a
+      # `C:\ProgramData\PuppetLabs\puppet\cache\client_data` directory with
+      # an owner of Administrator and a group of None. When the MCO service,
+      # running as SYSTEM, later triggers an on-demand Puppet run,
+      # a write to this directory fails with access denied since SYSTEM is not
+      # explicitly granted permissions. The catalog write occurs because MCO
+      # enables Puppet's cache terminus and therefore Puppet requires write
+      # permissions to store a JSON catalog.
+      #
+      # When MCO triggers the puppet run it will run as the SYSTEM user and will
+      # create the cache directory with an Owner of Administrators and a
+      # Group of SYSTEM.
+      windows_hosts.each do |h|
+        on(h , 'rm -rf "C:\ProgramData\PuppetLabs\puppet\cache\client_data"')
+      end
+
+    step "Run mco puppet runonce"
+      on(mco_master, "#{mco_bin} puppet runonce")
+      sleep 45
+
+    step "Verify that powershell created file on windows systems"
+      cmd = "cmd.exe /c \"dir c:\\#{testfile}.txt\""
+      windows_hosts.each do |h|
+        (1..10).each do |iter|
+          res = on(h, cmd, :acceptable_exit_codes => (0..254))
+          unless res.exit_code == 0
+            sleep (2**iter)/2
+          else
+            break
+          end
+        end
+        result = on(h, cmd, :acceptable_exit_codes => (0..254))
+        assert_equal(0, result.exit_code, 'mco failed to create file using powershell')
+      end
+  end
+
+end


acceptance/tests/mco_puppet_powershell.rb
```diff
@@ -0,0 +1,98 @@
...
+    if /windows/ =~ h[:platform]
+       windows_hosts << h
+    end
+  end
+  if windows_hosts.empty?
+    skip_test "No windows hosts to test powershell on"
+  end
+
+  step "Install powershell provider on the master" do
+    on master, puppet("module install puppetlabs/powershell")
```
Iristyle: @johnduarte I believe I've discovered your issue.

tl; dr - I believe this line should be:

```ruby
on master, puppet("module install puppetlabs-powershell")
```

### Steps that I used to verify

* First, I hopped on the Windows host, and manually triggered an agent run.  The result was:

```
Failed to apply catalog: Parameter provider failed on Exec[create-test-file]: Invalid exec provider 'powershell' at /tmp/mco_powershell.X0dZWk/environments/production/manifests/site.pp:3
```

* Verified that the agent did not have `powershell` provider in `C:\programdata\PuppetLabs\puppet\cache\lib\puppet\provider\exec\powershell.rb`

* Verified that the master didn't have powershell installed:

```
[root@gyzb8dxqu9uo8qt ~]# puppet module list

/etc/puppetlabs/code/modules (no modules installed)
/opt/puppetlabs/puppet/modules (no modules installed)
```

* Manually installed `powershell` on master with PMT and verified

```
[root@gyzb8dxqu9uo8qt ~]# puppet module install puppetlabs-powershell
Notice: Preparing to install into /tmp/mco_powershell.X0dZWk/environments/production/modules ...
Notice: Created target directory /tmp/mco_powershell.X0dZWk/environments/production/modules
Notice: Downloading from https://forgeapi.puppetlabs.com ...
Notice: Installing -- do not interrupt ...
/tmp/mco_powershell.X0dZWk/environments/production/modules
└── puppetlabs-powershell (v1.0.4)

[root@gyzb8dxqu9uo8qt ~]# puppet module list
/tmp/mco_powershell.X0dZWk/environments/production/modules
└── puppetlabs-powershell (v1.0.4)
/etc/puppetlabs/code/modules (no modules installed)
/opt/puppetlabs/puppet/modules (no modules installed)
```

* Re-ran the agent manually and watched it succeed locally, and create the appropriate text file.  Note that `C:\programdata\PuppetLabs\puppet\cache\lib\puppet\provider\exec\powershell.rb` exists now

* Delete the text file from `c:\mco_powershell.P0RJ9w.txt`

* Triggered mco run from master via `mco puppet runonce`

* Verified that `c:\mco_powershell.P0RJ9w.txt` is created 

### NOTE

* The event log still contains an error message, but this doesn't appear to be preventing MCO from doing it's job.  You might want to do some additional investigation to track down that error to determine if it's a new issue or not:

```
Could not retrieve catalog from remote server: No such file or directory @ dir_s_rmdir - C:/ProgramData/PuppetLabs/puppet/cache/client_data/catalog/nwpo2qqas4pd9j1.delivery.puppetlabs.net.json20150625-680-1l0wkf9.lock
```

This message originates from this point in Puppet - https://github.com/puppetlabs/puppet/blob/ab1ea395a4ca9333d43a0e64ed47d24c93a816e4/lib/puppet/configurer.rb#L300

* For fun, I tried to reproduce the original failure:
    * changed the plugin code in `puppet.rb` to use the previous process creation code `		   ::Process.spawn(command, :new_pgroup => true)`
    * restarted the MCO service on the agent
    * triggered another `mco puppet runonce` from the master

I couldn't reproduce the original crash.  It's late and I might have made a mistake here, but you might want to make sure this acceptance test indeed fails against the previous MCO code.  Also note, the `dir_s_rmdir` message appears in the event log as well here.


/cc @joshcooper @richardc

richardc: So in order to double-confirm I took this branch and changed the module install, and I see the powershell provider install on the master node, so that moves us forwards:

```
Begin tests/mco_puppet_powershell.rb

mco puppet run with powershell provider

iovr3k3i49l7cfm.delivery.puppetlabs.net (master) 12:08:37$ puppet module install puppetlabs-powershell
Notice: Preparing to install into /etc/puppetlabs/code/environments/production/modules ...
Notice: Downloading from https://forgeapi.puppetlabs.com ...
Notice: Installing -- do not interrupt ...
/etc/puppetlabs/code/environments/production/modules
└── puppetlabs-powershell (v1.0.4)

iovr3k3i49l7cfm.delivery.puppetlabs.net (master) executed in 1.57 seconds
```

But I'm still not seeing the powershell provider syncing:

```
Administrator@lwc8vbgcgmyifkq /cygdrive/c/ProgramData/PuppetLabs/puppet/cache/lib
$ find
.
```

And we still see this in the system error logs:

```
Could not retrieve catalog from remote server: No such file or directory @ dir_s_rmdir - C:/ProgramData/PuppetLabs/puppet/cache/client_data/catalog/lwc8vbgcgmyifkq.delivery.puppetlabs.net.json20150625-272-1trzzrb.lock
```

My current line of thinking is that by running `puppet agent -t` manually outside of mcollective that we're running with a different inherited environment and set of permissions that may allow puppet to change the permissions/ownerships over `C:/ProgramData/PuppetLabs/puppet/cache/client_data/catalog/` allowing for future MCollective hosted puppet runs to later spool a catalog and apply it, or re-apply the existing cached catalog.

Iristyle: Just to follow up on this set of comments - the catalog was cached with `puppet agent -t`, which was skewing my subsequent mco runs.

acceptance/tests/mco_puppet_powershell.rb
```diff
@@ -0,0 +1,98 @@
...
+    install_puppet_module_via_pmt_on(master, {:module_name => "puppetlabs-powershell"})
+  end
+
+  step "Ensure that agents have checked in with master" do
+    hosts.each do |host|
+      stub_hosts_on(host, 'puppet' => master.ip)
+    end
+    on master, puppet("resource service puppetserver ensure=running")
+    hosts.each do |h|
+      on h, puppet("agent -t")
```
Iristyle: So it's possible this is a source of our `dir_s_rmdir` problems.  If Puppet runs as a non-admin, when it creates and manages the MCO cache directory at `C:\ProgramData\PuppetLabs\puppet\cache\client_data`, it will strip the `SYSTEM` permissions.

The last notes from @richardc on the ticket seem to indicate the permissions are correct in his environment based on output from `icacls` at this point in time (see https://github.com/richardc/marionette-collective/commit/1620169bc4b2d88e2527366c2fd4ec0c98f57276), but from what I saw in the environment you prepared @johnduarte , the permissions were wrong.

A little more digging will be required to get to the bottom of this.  Also note that there is an issue filed about permission mangling in this non-admin scenario - see https://tickets.puppetlabs.com/browse/PUP-4684

richardc: I think that helper is running this command as Administrator, ssh-ing in.  Would Administrator-via-ssh and the LocalSystem version of Administrator have different effective groups?

Iristyle: One way to verify the user and privileges, is to execute a `whoami.exe /priv` prior to calling puppet.  The `puppet` beaker helper should just be a wrapper around this same type of command execution (https://github.com/puppetlabs/beaker/blob/dd70aa6644e5147140cd14562ef666d6345fd46e/lib/beaker/dsl/wrappers.rb#L57) so should be running as the same user as the `whoami` command.

That user *should* be the Cygwin service user, which is `cyg_server`.  Normally the Puppet service runs as `LocalSystem`.  So this could be the source of the permissions discrepancy.

richardc: This is the output from `whoami.exe /priv`:

```
o04edweo4pvgyua.delivery.puppetlabs.net (agent-2012r2-x86_64-rubyx64) 14:50:25$ whoami.exe /priv

PRIVILEGES INFORMATION
----------------------

Privilege Name                  Description                               State
=============================== ========================================= =======
SeIncreaseQuotaPrivilege        Adjust memory quotas for a process        Enabled
SeSecurityPrivilege             Manage auditing and security log          Enabled
SeTakeOwnershipPrivilege        Take ownership of files or other objects  Enabled
SeLoadDriverPrivilege           Load and unload device drivers            Enabled
SeSystemProfilePrivilege        Profile system performance                Enabled
SeSystemtimePrivilege           Change the system time                    Enabled
SeProfileSingleProcessPrivilege Profile single process                    Enabled
SeIncreaseBasePriorityPrivilege Increase scheduling priority              Enabled
SeCreatePagefilePrivilege       Create a pagefile                         Enabled
SeBackupPrivilege               Back up files and directories             Enabled
SeRestorePrivilege              Restore files and directories             Enabled
SeShutdownPrivilege             Shut down the system                      Enabled
SeDebugPrivilege                Debug programs                            Enabled
SeSystemEnvironmentPrivilege    Modify firmware environment values        Enabled
SeChangeNotifyPrivilege         Bypass traverse checking                  Enabled
SeRemoteShutdownPrivilege       Force shutdown from a remote system       Enabled
SeUndockPrivilege               Remove computer from docking station      Enabled
SeManageVolumePrivilege         Perform volume maintenance tasks          Enabled
SeImpersonatePrivilege          Impersonate a client after authentication Enabled
SeCreateGlobalPrivilege         Create global objects                     Enabled
SeIncreaseWorkingSetPrivilege   Increase a process working set            Enabled
SeTimeZonePrivilege             Change the time zone                      Enabled
SeCreateSymbolicLinkPrivilege   Create symbolic links                     Enabled
```

Iristyle: Also - I'm assuming that the reason we're running `agent` here is to induce a pluginsync and get MCO configured properly?

Is there another way of getting the same effect without calling `puppet agent`?

If we can think of an alternative mechanism for doing this, that might be another way of getting around this problem.  Some other ideas off the top of my head:

* use `psexec` to run the agent as `SYSTEM`
* manually create the directory structure with the appropriate perms (seems fraught with peril)
* delete the `C:\ProgramData\PuppetLabs\puppet\cache\client_data` directory after it's created, since we know that MCO will recreate when requested to run puppet by the master (the most straightforward solution, though hacky)

Also, one thing that I did not check was the behavior or running `puppet` as non-admin *after* MCO has run as admin and Puppet has run to create the directory structure.  I'm not sure if it tries to manage the directory and mangles the previously correct permissions or not.

richardc: MCO is already configured independently of puppet in this environment, so I'm not sure why the explicit `puppet agent -t` here - the step description is `"Ensure that agents have checked in with master"` but that was already established (if without a catalog run) [here](https://github.com/johnduarte/marionette-collective/blob/mcop-521/acceptance/setup/common/pre-suite/040_ValidateSignCert.rb#L6) when setting up the puppet client certificates.

Iristyle: Ah, thanks for pointing that out - I didn't know about `initialize_ssl`.  I do see what you're referring to in the code at https://github.com/puppetlabs/marionette-collective/blob/a185d3e1e141183886d42e7eda64052fd2da13dc/acceptance/lib/puppet/acceptance/common_utils.rb#L23

Maybe there's another reason @johnduarte did this (or maybe like me he didn't realize that it was being done already)?

So I think a really quick test here would be to add a 1-liner to delete `C:\ProgramData\PuppetLabs\puppet\cache\client_data` after calling `puppet agent -t`.

That should be sufficient to at least see if we can get tests passing.

johnduarte: @Iristyle , adding a step to delete the the `client_data` directory after `with_puppet_running_on` has been invoked allows the test to pass.

I will updated the PR to include this, remove the extra `puppet agent -t` call, and move the environment setup steps into the context of `with_puppet_running_on`. This has allowed the automated test to pass.

acceptance/tests/mco_puppet_powershell.rb
```diff
@@ -0,0 +1,106 @@
...
+      windows_hosts.each do |h|
+        on(h , 'rm -rf "C:\ProgramData\PuppetLabs\puppet\cache\client_data"')
+      end
+
+    step "Run mco puppet runonce"
+      on(mco_master, "#{mco_bin} puppet runonce")
+      sleep 45
+
+    step "Verify that powershell created file on windows systems"
+      cmd = "dir C:/#{testfile}.txt"
```
Iristyle: this should be `c:\`

johnduarte: I can't get the `C:\` to work. I am assuming that cygwin does not handle this properly?

Attempted with 

    cmd = "dir C:\\#{testfile}.txt"

joshcooper: `dir` is a shell built-in, so it needs to be `"cmd.exe /c dir c:\\#{testfile}.txt"`. Alternatively, can you use beaker's built-in `Host#file_exists?(path)`, so just 

```
assert_equal(true, host.file_exists?("C:\\#{testfile}.txt"))
```

Iristyle: Sorry for the lack of clarification there @johnduarte - thanks @joshcooper

acceptance/tests/mco_puppet_powershell.rb
```diff
@@ -0,0 +1,114 @@
...
+
+    step "Stub puppet and run agent to get certs"
+    hosts.each do |host|
+      stub_hosts_on(host, 'puppet' => master.ip)
+    end
+
+    step "Install powershell module on master"
+    install_puppet_module_via_pmt_on(master, {:module_name => "puppetlabs-powershell"})
+
+    step "Delete puppet cache dir on Windows hosts, cause `agent` run creatd it with wrong permissions"
```
Iristyle: nits

`cause` -> `because`
`creatd` -> `created`

acceptance/tests/mco_puppet_powershell.rb
```diff
@@ -0,0 +1,114 @@
...
+      # NOTE: `with_puppet_running_on` triggers a `puppet agent -t` run on the
+      # Windows agents using the ADMINISTRATOR user. This creates a
+      # `C:\ProgramData\PuppetLabs\puppet\cache\client_data` directory with no
+      # group permissions. This will prevent the catalog from being saved when
+      # the puppet run in triggered by MCO.
+      #
+      # When MCO triggers the puppet run it will run as the SYSTEM user and will
+      # create the cache directory with SYSTEM group permissions.
```
Iristyle: `will create the cache directory with an owner of Administrators and a Group of SYSTEM.`

acceptance/tests/mco_puppet_powershell.rb
```diff
@@ -0,0 +1,114 @@
...
+      stub_hosts_on(host, 'puppet' => master.ip)
+    end
+
+    step "Install powershell module on master"
+    install_puppet_module_via_pmt_on(master, {:module_name => "puppetlabs-powershell"})
+
+    step "Delete puppet cache dir on Windows hosts, cause `agent` run creatd it with wrong permissions"
+      # NOTE: `with_puppet_running_on` triggers a `puppet agent -t` run on the
+      # Windows agents using the ADMINISTRATOR user. This creates a
+      # `C:\ProgramData\PuppetLabs\puppet\cache\client_data` directory with no
```
Iristyle: directory with an owner of Administrator and a group of None.  When the MCO service, running as SYSTEM, later triggers an on-demand Puppet run, a write to this directory fails with access denied since SYSTEM is not explicitly granted permissions.  The catalog write occurs because MCO enables Puppet's cache terminus and therefore Puppet requires write permissions to store a JSON catalog.

richardc:
Good to see we understand the testing disparity and have a workaround.

@johnduarte  I'm happy to merge this once the nits @Iristyle has noted are fixed up,  to save on splinters from head-scratching next time.

johnduarte:
Thank you @Iristyle for clarifying the details of the issue. I have amended this commit to include your comments.

acceptance/tests/mco_puppet_powershell.rb
```diff
@@ -0,0 +1,119 @@
...
+  master_opts = {
+    'main' => {
+      'environmentpath' => "#{testdir}/environments",
+     }
+  }
+
+  with_puppet_running_on(master, master_opts) do
+    if mco_master.platform =~ /windows/ then
+      if mco_master[:ruby_arch] == 'x86' then
+        mco_bin = 'cmd.exe /c "C:\Program Files (x86)\Puppet Labs\Puppet\bin\mco.bat"'
```
joshcooper: I think this path only makes sense when `ruby_arch` is x86, but `platform` is x64. If both are x86 or both are x64, then you'll want to use the path below `'C:\Program Files\Puppet Labs\...'`

acceptance/tests/mco_puppet_powershell.rb
```diff
@@ -0,0 +1,119 @@
...
+      # a write to this directory fails with access denied since SYSTEM is not
+      # explicitly granted permissions. The catalog write occurs because MCO
+      # enables Puppet's cache terminus and therefore Puppet requires write
+      # permissions to store a JSON catalog.
+      #
+      # When MCO triggers the puppet run it will run as the SYSTEM user and will
+      # create the cache directory with an Owner of Administrators and a
+      # Group of SYSTEM.
+      windows_hosts.each do |h|
+        on(h , 'rm -rf "C:\ProgramData\PuppetLabs\puppet\cache\client_data"')
```
joshcooper: you can ask the host for the path using `puppet agent --configprint client_datadir`

Iristyle: Shoot - I meant to mention that as well.

johnduarte: Thanks @joshcooper . I will submit a MAINT PR to address these issues.

acceptance/tests/mco_puppet_powershell.rb
Iristyle: Also agreed here.

PR #323, opened 2015-06-25
(MAINT) fix puppet bin location on win 32bit

johnduarte:
This commit fixes the puppet command path on Windows 32bit
architectures in the 060_Install-activemq.rb acceptance test
pre-suite.


acceptance/setup/aio/pre-suite/060_Install-activemq.rb
```diff
@@ -122,8 +122,12 @@
...
       libdir = "#{mco_confdir}/plugins"
       logdir = "C:/ProgramData/PuppetLabs/mcollective/var/log"
-      puppet_cmd = "C:/Program Files/Puppet Labs\\Puppet\\bin\\puppet.bat"
-      puppet_confdir = "C:/ProgramData/PuppetLabs\\puppept\\etc\\puppet"
+      if h[:ruby_arch] == 'x64' then
+        puppet_cmd = "C:/Program Files/Puppet Labs/Puppet/bin/puppet.bat"
+      else
+        puppet_cmd = "C:/Program Files (x86)/Puppet Labs/Puppet/bin/puppet.bat"
+      end
+      puppet_confdir = "C:/ProgramData/PuppetLabs/puppept/etc/puppet"
```
richardc: That appears to be a typo - `/puppept/etc`

richardc:
One thing that really strikes me as odd with some reflection, why is this in 060_install_activemq.rb?  It's not about installing activemq, it's about configuring the mcollective daemon on this node.

Also as this is here, should we delete this: https://github.com/puppetlabs/marionette-collective/blob/master/acceptance/files/windows-server.cfg

johnduarte:
Thank you for your feedback @richardc . I have updated the PR to reflect your suggestions.

richardc:
This still does both:
https://github.com/johnduarte/marionette-collective/blob/fix_x86_puppet_bin/acceptance/setup/aio/pre-suite/060_Install-mcollective-daemon.rb

It installs activemq in lines 1-105, and 192-195 it verifies that it is running.  Then it configures mcollective in lines 106..191.

These aren't the same component.  If we want to install both in the one fixture it should be called 'install activemq and configure mcollective server'  but I think it'd be less weird to pull them further apart.

johnduarte:
@richardc I have split this pre-suite into two components:
`050_Install-activemq.rb` and
`060_Install-mcollective-daemon.rb`

richardc:
This test is still out of place - https://github.com/johnduarte/marionette-collective/blob/fix_x86_puppet_bin/acceptance/setup/aio/pre-suite/060_Install-mcollective-daemon.rb#L90-L93

It's verifying that activemq is listening on port 61613, but it's in the test `configure mcollective daemon`

PR #327, opened 2015-07-15
(MAINT) Fixup Windows elements in powershell test

johnduarte:
This commit replaces the hard-coded client_datadir with a puppet
configprint for this value.

This commit also corrects a logical error in the arch and ruby_arch
combination that would result in the mco bindir being located under
"C:\Program Files (x86)".


acceptance/tests/mco_puppet_powershell.rb
```diff
@@ -61,7 +61,7 @@
 
   with_puppet_running_on(master, master_opts) do
     if mco_master.platform =~ /windows/ then
-      if mco_master[:ruby_arch] == 'x86' then
+      if mco_master[:ruby_arch] == 'x86' and mco_master.platform.arch == '64' then
```
joshcooper: Is `mco_master.platform.arch` a new beaker thing? I'm used to seeing `mco_master['platform'] =~ /x64/`

acceptance/tests/mco_puppet_powershell.rb
```diff
@@ -93,7 +93,8 @@
       # create the cache directory with an Owner of Administrators and a
       # Group of SYSTEM.
       windows_hosts.each do |h|
-        on(h , 'rm -rf "C:\ProgramData\PuppetLabs\puppet\cache\client_data"')
+        client_datadir = on(h, puppet("agent --configprint client_datadir")).stdout.chomp
```
joshcooper: I think you can do `client_datadir = h.puppet['client_datadir']` and it will do the stdout.chomp business. Also it defaults to `agent` run mode, though if you needed to you can specify the run mode explicitly `h.puppet('master')['some_setting']`

acceptance/tests/mco_puppet_powershell.rb
johnduarte: Yes, beaker provides a platform object that contains the arch, codename, variant, and version.

http://www.rubydoc.info/github/puppetlabs/beaker/master/Beaker/Platform

Iristyle: It's too bad we can't just use `puppetbindir` to find `mco.bat` rather than all this branching logic.  You would have to split on the pathseparator / test each path ... but it would be less hardcoding.

https://github.com/puppetlabs/beaker/blob/11f9a9bae061dec0b81c3122bed19e6449d7104c/lib/beaker/dsl/install_utils/foss_defaults.rb#L82

https://github.com/puppetlabs/beaker/blob/11f9a9bae061dec0b81c3122bed19e6449d7104c/lib/beaker/dsl/install_utils/aio_defaults.rb#L19

https://github.com/puppetlabs/beaker/blob/11f9a9bae061dec0b81c3122bed19e6449d7104c/lib/beaker/dsl/install_utils/pe_defaults.rb#L52

johnduarte: Oh, if the `puppetbindir` is already on the path, then we probably do not need any of this. Just use

    mco_bin = 'cmd.exe /c "mco.bat"'
    on(mco_master, "#{mco_bin} puppet runonce")

joshcooper: yeah, that's way better, :+1:

minor nit, you can drop the inner quotes around mco.bat.

Btw, if for some reason you also need `privatebindir` then you can do:

```
 "env PATH=\"#{host['privatebindir']}:${PATH}\" cmd /c mco"
```

like we do for executing gem commands: https://github.com/puppetlabs/puppet/blob/master/acceptance/lib/puppet/acceptance/common_utils.rb#L136

johnduarte:
Thank you @Iristyle for pointing out the opportunity to remove the arch logic. I have also loaded the fact resolution component that we discussed into the powershell test.

acceptance/tests/mco_puppet_powershell.rb
```diff
@@ -113,6 +110,8 @@
         end
         result = on(h, cmd, :acceptable_exit_codes => (0..254))
         assert_equal(0, result.exit_code, 'mco failed to create file using powershell')
+        res = on h, "cat 'C:/#{testfile}.txt'"
+        assert_match(/#{h}/, res.stdout, "FQDN fact not included in file generated by mco")
```
joshcooper: I'm not following how `/#{h}/` will match `res.stdout`? Does `Host#to_s` return the fqdn, if so, can we make that more explicit, e.g. `assert_match(h.hostname, res.stdout, ...)`

acceptance/tests/mco_ping.rb
```diff
@@ -1,13 +1,9 @@
 test_name "mco ping" do
   step "run ping"
     if mco_master.platform =~ /windows/ then
-      if mco_master[:ruby_arch] == 'x86' then
-        mco_bin = 'cmd.exe /c C:/Program\ Files\ \(x86\)/Puppet\ Labs/Puppet/bin/mco.bat'
-      else
-        mco_bin = 'cmd.exe /c C:/Program\ Files/Puppet\ Labs/Puppet/bin/mco.bat'
-      end
+      mco_bin = 'cmd.exe /c mco.bat'
```
Iristyle: :+1: much better

acceptance/tests/mco_puppet_powershell.rb
Iristyle: agreed

PR #328, opened 2015-07-17
(RE-5032) fully qualify the label in the osx mco plist

heathseals:
This commit fully qualifies the label in the osx mco plist using the
identifier plus service name.

### DIFF
--- a/ext/aio/osx/mcollective.plist
+++ b/ext/aio/osx/mcollective.plist
@@ -10,7 +10,7 @@
                 <string>en_US.UTF-8</string>
         </dict>
         <key>Label</key>
-        <string>mcollective</string>
+        <string>com.puppetlabs.mcollective</string>
         <key>KeepAlive</key>
         <true/>
         <key>ProgramArguments</key>


joshcooper:
puppet-agent currently ships with [2.8.2](https://github.com/puppetlabs/puppet-agent/blob/aardwolf/configs/components/marionette-collective.json#L3), should this be backported to 2.8.x or stable (I'm not sure if/how those are different)?

richardc:
@joshcooper until your comment nobody had expressed the need for any form of timely release with this change incorporated and we currently have no releases planned.

What time pressure do we have to get this released?

joshcooper:
@richardc we're planning on releasing OSX in PE real soon now, and if there are simple things we can do to improve the OSX experience that'd be great. But at the same time, it's not worth a new mcollective release just for this one thing. So I'd say if we are planning an mco release to coincide with the next PE .z release that'd be great, but otherwise let's wait.

Btw, we made a similar change in to puppet in https://github.com/puppetlabs/puppet/pull/4095

PR #330, opened 2015-08-03
Adding useful tip for managing Java Heap at runtime.

dforste:


richardc:
How would a user know if they were using the default bin/activemq script?

dforste:
If they download the tar install and run bin/activemq . . .

richardc:
OK, so where in the rest of this document do we tell the user to do that?

dforste:
Good point perhaps should change to if using a source install you can do x.

richardc:
We'd also need to start to explain to the user how to do a source install, which would be out of scope for what this document is intended to communicate.

If you want to rework this more so that it fits then please do, but as it stands this isn't a good addition to the document.

PR #331, opened 2015-08-05
(MCO-687) Add solaris smf service for mcollective for AIO

haus:
This commit adds an smf service file for mcollective, based on the PE
service file with updates for the AIO FS layout.


shrug:
:+1:

PR #332, opened 2015-08-14
update links to fuse docs - moved under access.redhat.com

jessereynolds:
The Fuse docs have all moved from the fusesource.com domain to be under access.redhat.com. The actual pages seem to have moved around a bit as well. This PR updates the links to the fuse docs. 

It also updates the main URL for mcollective in the README to be http://puppetlabs.com/mcollective as it appears http://marionette-collective.org/ is no longer (or it's temporarily offline perhaps?)


richardc:
Thanks.

PR #335, opened 2015-08-20
(MCO-705) revert "(RE-5032) fully qualify the label in the osx mco pl…

MikaelSmith:
…ist"

Reverts puppetlabs/marionette-collective#328

We realized that changing this name actually changes the service name,
which impacts anyone trying to manage the service - such as PE. This
seems high-impact for a patch release.


joshcooper:
:+1:

PR #337, opened 2015-08-21
(MCO-760) Runtime username/password input to connect to middleware.

devghai:
Adding support to interactively add username and/or password to connect to activemq and/or rabbitmq, if plugin.activemq.prompt_user and plugin.activemq.prompt_password are set to true respectively. This allows for safer deployment of MCollective when RabbitMQ/ActiveMQ is integrated with LDAP. In absence of this facility, LDAP passwords had to be stored in cleartext in conf file in user's home directory. Username and password entered on command line will take precedence over the ones in conf file, even if they are set, if the respective flags are set to true.

sample ~/.mcollective

...
connector = activemq
**plugin.activemq.prompt_password = true**
**plugin.activemq.prompt_user = false**
plugin.activemq.pool.size = 2
plugin.activemq.pool.1.host = stomp1
plugin.activemq.pool.1.port = 6163
plugin.activemq.pool.1.user = mcollective
plugin.activemq.pool.2.host = stomp_fallback
plugin.activemq.pool.2.port = 6163
plugin.activemq.pool.2.user = fallback
...

First time ever with Ruby.

----
Edit log: Changing pull request comments as the config setting that enables this has been changed from original idea.


lib/mcollective/monkey_patches.rb
```diff
@@ -148,3 +148,11 @@ def initialize(*args)
     set_params(params)
   end
 end
+
+unless Kernel.respond_to?(:require_relative)
```
richardc: No, this is terrible, please don't do this.

lib/mcollective/connector/password.rb
```diff
@@ -0,0 +1,158 @@
...
+#   OR  PROFITS;  OR  BUSINESS INTERRUPTION)  HOWEVER  CAUSED  AND  ON  ANY  THEORY OF 
+#   LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR 
+#   OTHERWISE) ARISING IN ANY WAY OUT OF THE  USE OF THIS SOFTWARE, EVEN IF ADVISED OF
+#   THE POSSIBILITY OF SUCH DAMAGE.
+
+#   Some modifications by Dev Ghai (devghai@gmail.com) to fit in with MCollective.
+ 
+ # Ask for a password from CLI, 
+# if it exists.
+module Password
```
richardc: If we want to keep this at all it should be in MCollective::Util::PasswordEntry or something.  It's not anything really to do with connectors, so it being in liib/mcollective/connectors/password.rb is just the wrong place.

lib/mcollective/connector/activemq.rb
```diff
@@ -1,4 +1,5 @@
 require 'stomp'
+require_relative 'password'
```
richardc: This shouldn't be a relative require, because it's not sane.   After moving it would need to be a `require mcollective/util/password_prompter` or whatever

lib/mcollective/connector/activemq.rb
```diff
@@ -234,6 +237,41 @@ def connect(connector = ::Stomp::Connection)
             host[:passcode] = get_env_or_option("STOMP_PASSWORD", "activemq.pool.#{poolnum}.password", '')
             host[:ssl] = get_bool_option("activemq.pool.#{poolnum}.ssl", "false")
 
+            if host[:login] == 'ask'
```
richardc: I don't care for this overriding of the parameter value.   It would be better to add an additional parameter to the plugin namespace. `plugin.activemq.interactive-user-prompt` (but with a better name), which would default false.

richardc:
This is an interesting idea, but it needs a bunch of tests to be added, and a bunch of refactoring to take some of the weirdness out of it.

devghai:
Apologies as there were other tasks that needed my attention.

All comments are appreciated! will work on them.

lib/mcollective/monkey_patches.rb
devghai: What's the Ruby way to do this? This is the best approach I could find online that works with all versions of Ruby that Travis-CI compiles MCollective against. Any pointers to good examples would be welcome.

lib/mcollective/connector/password.rb
devghai: I thought the same thing, but wanted to bang out the first pass real quick. I need to read code again to figure out how are modules and their dependencies loaded.

lib/mcollective/connector/activemq.rb
devghai: plugin.activemq.pool.1.prompt-user = [true|false, default=false]
plugin.activemq.pool.1.prompt-password = [true|false, default=false]
plugin.rabbitmq.pool.1.prompt-user = [true|false, default=false]
plugin.rabbitmq.pool.1.prompt-password = [true|false, default=false]

Do these suggestions look good?

lib/mcollective/monkey_patches.rb
richardc: It the ruby way of doing it, but I don't want to carry a monkey patch for a utility of questionable value.

I don't want our code to use relative_require, see the comments on the namespace of lib/mcollective/connector/password.rb for more on why.

lib/mcollective/connector/password.rb
```diff
@@ -0,0 +1,158 @@
+#   Copyright 2012-2013 Barry Allard <barry.allard@gmail.com>
+#   All rights reserved.
```
richardc: I'm not sure we can take this, as it's adding additional copyright and advertising burden on us.

devghai:
I am writing test case for this request and I noticed middleware passwords for 2 pools are different. This change assumes all pools have same password. I am changing the prompt per pool considering that. Caveat of this is that if multiple pools are configured to prompt for password, it would have to be entered number_of_pool times for every command... which would be quite a bad user experience. Alternative would be have a flag like plugin.rabbitmq.all_pools_have_same_password.

Thoughts?

lib/mcollective/connector/activemq.rb
```diff
@@ -224,15 +224,43 @@ def connect(connector = ::Stomp::Connection)
 
           pools = Integer(get_option("activemq.pool.size"))
           hosts = []
+          middleware_user = ''
+          middleware_password = ''
+          prompt_for_username = get_bool_option("activemq.prompt_user", "false")
+          prompt_for_password = get_bool_option("activemq.prompt_password", "false")
+          
+          if prompt_for_username and middleware_user == ''
```
richardc: At this point it will always be '', because you didn't look to the settings for a configured value.

lib/mcollective/connector/activemq.rb
```diff
@@ -224,15 +224,43 @@ def connect(connector = ::Stomp::Connection)
...
 
             host[:host] = get_option("activemq.pool.#{poolnum}.host")
             host[:port] = get_option("activemq.pool.#{poolnum}.port", 61613).to_i
-            host[:login] = get_env_or_option("STOMP_USER", "activemq.pool.#{poolnum}.user", '')
-            host[:passcode] = get_env_or_option("STOMP_PASSWORD", "activemq.pool.#{poolnum}.password", '')
             host[:ssl] = get_bool_option("activemq.pool.#{poolnum}.ssl", "false")
+            
+            # read user from config file if activemq user is not set for prompt
+            unless prompt_for_username
+              host[:login] = get_env_or_option("STOMP_USER", "activemq.pool.#{poolnum}.user", '')
```
richardc: This can be far simpler.  Remove all the branching and put it back to where it was, but use middleware_user as the default value.  

```
host[:login] = get_env_or_option("STOMP_USER", "activemq.pool.#{poolnum}.user", middleware_user)
```

richardc:
I think it's find that this setting would be connector rather than pool member level, as it's just a fancier way of setting `STOMP_USER` and `STOMP_PASSWORD`

lib/mcollective/connector/activemq.rb
```diff
@@ -1,4 +1,4 @@
-require 'stomp'
+require 'stomp'
```
richardc: Fix this whitespace.

lib/mcollective/connector/activemq.rb
devghai: true. `and middleware_user == ''` is redundant.

lib/mcollective/connector/activemq.rb
devghai: The suggestion would mean the user input will be ignored even if the flag is set, if user/password is entered in config file for a pool. This will lead to unnecessary typing and (maybe) confusion for a user (if password in conf file is wrong while the user is entering the right password).

richardc: So you're saying that this wouldn't be confusing to a user?

```
plugin.activemq.prompt_user = true
plugin.activemq.pool.1.user=nothing_that_will_be_used
```

I suspect the connector should just explode if the user did that, but the code will need to look twice to find out if they did it.

devghai: Well... that's why the original suggestion of 'ask' for that value... so as to clearly distinguish between behaviour at runtime. In my opinion, because we are puppetizing mco deployment and end users are going to be developers who may not be going around configs too much, silent rejection of entered password at runtime might be not be good user experience.

Left to me, I would go back to 'ask' approach. I specially chose that word because it is too small to be a legitimate password in any shop, in addition to clearly changing runtime behaviour. Even default RabbitMQ password for default user `guest` is `guest`. On the flip-side, because this approach of flag is not followed anywhere else in the configuration, it will not be as intuitive.

Considering you have more experience with MCollective and it's deployment and use, please suggest an approach and I will refactor accordingly. I initially thought of setting the default value to middleware_password/middleware_user too, but did not do it because of the reasons outlined above.

Thank you for the time you are spending on this!

Dev

devghai:
Please let me know if everything is OK or there are some changes that you'd like to see to get the pull request merged in.

fabianmenges:
Is this going to get merged in? Would be pretty useful.

lib/mcollective/util.rb
```diff
@@ -520,5 +520,60 @@ def self.field_number(field_size, max_size=90)
...
     end
+    
+    def self.get_hidden_input_on_windows()
+      require 'Win32API'
+      # Hook into getch from crtdll. Keep reading all keys till return
+      # or newline is hit.
+      # If key is backspace or delete, then delete the character and update
+      # the buffer.
+      input = ''
+      while char = Win32API.new("crtdll", "_getch", [ ], "I").Call do
```
MikaelSmith: This is potentially generating a lot of objects. Better to init the object for _getch once, then call it repeatedly.

MikaelSmith:
I'd like to see specs around enabling the new options. That might require mocking `get_hidden_input`, since those are trickier to spec test. https://github.com/puppetlabs/marionette-collective/blob/master/spec/unit/mcollective/discovery/stdin_spec.rb might have some useful patterns.

MikaelSmith:
Was there ever an MCO ticket created for this?

devghai:
i don't know so. i needed it, so made it and contributed it upstream. we should probably look if there is any associated ticket and close it. can do that once i get back from vacation.

MikaelSmith:
The ticket will be reviewed Wednesday morning. We also have a Pull Request Triage for core projects next Tuesday (May 17th, 10am PDT), if it doesn't get merged before then.

MikaelSmith:
Are you able to add spec tests around the enabling the new options? Currently they all run with it disabled.

devghai:
I should be able to do that, just not able to get to it because of 
current workload. Sorry :(

MikaelSmith:
I think I'll merge it. If you have time in the future please add some extra tests.

PR #343, opened 2015-09-04
(RE-5539) Set SSL environment variables to point to secure locations

melissa:


melissa:
Hmm, this is potentially not necessary, though I'm not positive. More investigations in progress

melissa:
Yeah, I think this isn't needed. Will reopen if something indicates otherwise

PR #345, opened 2015-09-12
(maint) Fix Solaris SMF service for mcollective for AIO

geoffnichols:
Prior to this commit, mcollective service startup
with the Solaris AIO puppet-agent fails because the
SMF manifest is using the non-AIO path to mcollectived.

This commit updates the Solaris SMF service manifest
to use the correct AIO path to mcollectived.

### DIFF
--- a/ext/aio/solaris/smf/mcollective.xml
+++ b/ext/aio/solaris/smf/mcollective.xml
@@ -24,7 +24,7 @@
       <service_fmri value="svc:/system/filesystem/local"/>
     </dependency>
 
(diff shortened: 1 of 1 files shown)


stahnma:
:+1:

kylog:
:+1: 

While we're poking at wrong paths, https://github.com/puppetlabs/marionette-collective/blob/master/ext/aio/redhat/mcollective.sysconfig#L4-L7 is wrong as well. I don't know if this could result in any breakage since it's all commented out (but perhaps something uses it as a template and uncomments things?).

PR #348, opened 2015-10-27
(docs) Remove Chef docs from the Puppet docs.

gguillotte:


nfagerlund:
Checked with @richardc, and we're good to go.

PR #349, opened 2015-10-27
(docs) Update or fix links, especially to plugins.

gguillotte:
Many links point to the projects.puppetlabs.com wiki, instead of the
migrated equivalent on docs.puppetlabs.com. Point these links to the
best places.

Also, several links point to paths on marionette-collective.org. Point
these links to docs.puppetlabs.com/mcollective.


gguillotte:
Rebased.

PR #350, opened 2015-10-27
(maint) Remove Fedora 20 from build targets

melissa:
Fedora 20 is EOL, so we shouldn't be building packages for it.


melissa:
I can't remember if we're building packages or not, but just to be safe, here's this

PR #353, opened 2015-11-04
(docs) Correct AES Security Plugin title

alexjfisher:
I noticed that both security_ssl.md and security_aes.md had the same
page title.

The aes security plugin is now filed under 'Older and
Non-Recommended Information'
https://github.com/puppetlabs/puppet-docs/blob/master/source/_includes/mcollective_menu.html#L89

As such, I've updated the AES page to more strongly suggest the use of
the SSL plugin.

### DIFF
--- a/website/reference/plugins/security_aes.md
+++ b/website/reference/plugins/security_aes.md
@@ -1,6 +1,6 @@
 ---
 layout: default
-title: OpenSSL based Security Plugin
+title: AES Security Plugin
 ---
 [SimpleRPCAuthorization]: /mcollective/simplerpc/authorization.html
 [Registration]: registration.html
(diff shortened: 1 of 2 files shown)


richardc:
I'm :+1: for the title changes, but less so for the suggestions about whether to use the provider - the wording doesn't feel quite in-style for the rest of the page.  Ping @gguillotte and @nfagerlund for a docs eye on that.

alexjfisher:
Fair enough, I've used 'probably' twice.  I wasn't confident enough to say that the plugin was deprecated (although it does come under the 'Older and Non-Recommended Information' section).

gguillotte:
I'm :+1: for this PR. There are typos and style issues across this article, so I'm not fussed about the wording; the whole thing needs an editing pass anyway.

richardc:
Cool. Merging as-is then.  Thanks.

PR #354, opened 2015-11-25
Fix incorrect test for nil function return

abelletti:
The existing code wasn't a valid test for nil.  It would prematurely exit if the function being evaluated returned false.  This meant that a test like "provision().disabled=false" would never succeed even if provision().disabled evaluated to false.

Hit this when upgrading from MCO 2.2.4 to 2.8.4.

### DIFF
--- a/lib/mcollective/matcher.rb
+++ b/lib/mcollective/matcher.rb
@@ -136,7 +136,7 @@ module MCollective
       l_compare = execute_function(function_hash)
 
       # Break out early and return false if the function returns nil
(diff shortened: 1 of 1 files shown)


richardc:
Can you add a unit test so that this doesn't regress?

richardc:
Superceded by #355 which adds tests and further cleanup of this code.

abelletti:
Just returned from family visit -- sorry for the non-response and thanks for the fix!

PR #357, opened 2015-12-16
(MCO-742) Fix MCollective::RPC#run on Windows

reidmv:
systemu needs to be able to figure out a path to a working Ruby executable in order to work correctly. Because the systemu defaults don't reliably work as written when paths contain spaces, change the invocation of system() from single string to individual argument style invocation to clearly delineate command from argument without relying on system()'s parsing.

### DIFF
--- a/lib/mcollective/vendor/systemu/lib/systemu.rb
+++ b/lib/mcollective/vendor/systemu/lib/systemu.rb
@@ -28,10 +28,10 @@ class SystemUniversal
 
   c = begin; ::RbConfig::CONFIG; rescue NameError; ::Config::CONFIG; end
   ruby = File.join(c['bindir'], c['ruby_install_name']) << c['EXEEXT']
-  @ruby = if system('%s -e 42' % ruby)
+  @ruby = if system(ruby, '-e', '42')
     ruby
   else
-    system('%s -e 42' % 'ruby') ? 'ruby' : warn('no ruby in PATH/CONFIG')
+    system('ruby', '-e', '42') ? 'ruby' : warn('no ruby in PATH/CONFIG')
   end
 
   class << SystemUniversal


reidmv:
The Travis failures look to me like they might be more related to Travis' infrastructure migration ("new platform for Legacy Precise builds") than to the actual tests.

richardc:
Looks good, but plain to wait on the upstream duplicate (https://github.com/ahoward/systemu/pull/42)  being merged so we can be sure this won't regress if/when we unvendor systemu.

kylog:
Gonna try a close and re-open to poke travis.

kylog:
Also, given that we'd like to get this out the door, that we have no current plans to unvendor systemu, and the upstream is on a yak-shaving expedition, I'd like to get this merged. Perhaps we can just add a README_BEFORE_UNVENDORING.md referencing the upstream issue, and call it a day.

kylog:
@Iristyle or @ferventcoder can one of you review this? We'd like to get this in for the upcoming release. We were hoping for some independent validation from upstream, but if that doesn't happen in the next day or so, it would be nice to get independent review from Windows SMEs here.

lib/mcollective/vendor/systemu/lib/systemu.rb
```diff
@@ -28,10 +28,10 @@ def version() SystemUniversal::VERSION end
 
   c = begin; ::RbConfig::CONFIG; rescue NameError; ::Config::CONFIG; end
   ruby = File.join(c['bindir'], c['ruby_install_name']) << c['EXEEXT']
-  @ruby = if system('%s -e 42' % ruby)
+  @ruby = if system(ruby, '-e', '42')
     ruby
   else
-    system('%s -e 42' % 'ruby') ? 'ruby' : warn('no ruby in PATH/CONFIG')
+    system('ruby', '-e', '42') ? 'ruby' : warn('no ruby in PATH/CONFIG')
```
ferventcoder: I'm not even sure what this means :smile_cat: - if it does what the commit message says it does though, then :+1:

ferventcoder: Looks like it is a check to see if Ruby is available?

Iristyle: The difference here for the first line change is that `'%s -e 42' % ruby` yields the string value `"C:/Program Files/Puppet Labs/Puppet/sys/ruby/bin/ruby.exe -e 42"`, while `ruby, '-e', '42'` is a list of args, where the first arg is a path that includes spaces.

The docs for system state there are a few overloads for `system([env,] command... [,options]) → true, false or nil`
http://ruby-doc.org/core-2.2.0/Kernel.html#method-i-system

Implementation is in `process.c` at https://github.com/ruby/ruby/blob/trunk/process.c#L3990-L4021

To really understand the params, you have to look at `Process.spawn` docs at https://github.com/ruby/ruby/blob/trunk/process.c#L4059-L4325

In any event, in an irb session under `Puppet command prompt`, I get the following:

```
irb(main):007:0> system(ruby, '-e', '42')
=> true
irb(main):008:0> system('%s -e 42' % ruby)
=> nil
```

This is correct, though I'd like to see the commit message contain a bit more detail on this point to clear up why the patch actually works.

The change at line 34 shouldn't actually be necessary since the path in this case is simply `ruby` and contains no spaces that would confuse the `system` call (but is fine for the sake of consistency):  

```
irb(main):009:0> system('ruby', '-e', '42')
=> true
irb(main):010:0> system('ruby -e 42')
=> true
```


Overall :+1: with some commit message cleanup.

ferventcoder: I also ran irb to gain an understanding of what it does. :)

reidmv:
@ferventcoder I updated the commit message with a bunch of detail. Let me know if there's anything specific you'd like me to add; I didn't include the source references, just functional commentary on why it's desirable to invoke system() like this in this instance.

ferventcoder:
@puppetlabs/windows @Iristyle thoughts?

richardc:
In rewording the commit message to be more educational, it's now wrong.

```
arguments to the executable. The call system("/bin/echo one two") would
be broken down to an argument vector of ["/bin/echo", "one", "two"]
because Ruby would split it on spaces.
```

Ruby would not be the thing performing this tokenisation.  It's the system shell that performs the parsing.

richardc:
Has been cherry-picked and merged for the 2.8.7 release as ba8f93c.  Closing this PR as it's now redunant.

Thanks.

PR #360, opened 2016-01-07
(MCO-550) Use RubyW for service 2.7.x backport

ferventcoder:
Previously, any powershell related functions (facter facts, providers, execs,
MCO triggered powershell commands) fail to return 0 properly, and therefore
fail, but only when triggered via the MCO daemon under windows.

Switching back to using rubyw now that bug #7239 was backported to Ruby
1.9.3 (https://bugs.ruby-lang.org/issues/9232). This allows us to take
advantage of using rubyw and providing proper exit codes for powershell
processes.

This is a backport of 36bb928b64b47928dbfbc7308aa818d8250ee9ba, which is included in 2.8.x+

### DIFF
--- a/ext/windows/daemon.bat
+++ b/ext/windows/daemon.bat
@@ -3,4 +3,4 @@ SETLOCAL
 
 call "%~dp0..\bin\environment.bat" %0 %*
 
-ruby -rubygems "%MCOLLECTIVE_DIR%\bin\mcollectived" %*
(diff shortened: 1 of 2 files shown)


kylog:
No one is asking for a FOSS 2.7.1 release, so I'm closing this. We can re-open if there's demand.

ferventcoder:
:+1:

PR #361, opened 2016-01-13
WIP - 2.8.7 Release preparations

richardc:
This is the first of two possible 2.8.7 releases (the other being #362).  

This does not incorporate the changes from #357 and just prepares the fixes currently available on master for 2.8.7.


richardc:
We went with the more agressive version.

PR #363, opened 2016-01-21
(MCO-745) Add windows support to install.rb

melissa:
We need to be able to support installing Marionette Collective with the
provided install.rb file on windows. This commit does that.

We need to avoid running File.join on our paths on windows, because it
likes to add a `/` to the beginning of the path, so we end up with
`/C:/`, which throws an error.

### DIFF
--- a/install.rb
+++ b/install.rb
@@ -46,8 +46,7 @@ rescue LoadError
 end
 
 if (defined?(RbConfig) ? RbConfig : Config)::CONFIG['host_os'] =~ /mswin|win32|dos|mingw|cygwin/i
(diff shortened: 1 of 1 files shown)


install.rb
```diff
@@ -46,8 +46,7 @@
 end
 
 if (defined?(RbConfig) ? RbConfig : Config)::CONFIG['host_os'] =~ /mswin|win32|dos|mingw|cygwin/i
-    $stderr.puts "install.rb does not support Microsoft Windows. See ext/windows/README.md for information on installing on Microsoft Windows."
-    exit(-1)
+  WINDOWS = TRUE
 end
```
richardc: This is going to need a:

```
else
  WINDOWS = FALSE
end
```

richardc:
Have prepared #364 as a fixup commit.

PR #364, opened 2016-01-22
(MCO-745) Unbreak install.rb for non-windows platforms

richardc:
Here we define the WINDOWS constant to be FALSE on platforms that it is false
for, so we don't break there.

    install.rb:302:in `block in <main>': uninitialized constant WINDOWS (NameError)
            from /opt/puppetlabs/puppet/lib/ruby/2.1.0/fileutils.rb:125:in `chdir'
            from /opt/puppetlabs/puppet/lib/ruby/2.1.0/fileutils.rb:125:in `cd'
            from install.rb:293:in `<main>'

### DIFF
--- a/install.rb
+++ b/install.rb
@@ -47,6 +47,8 @@ end
 
 if (defined?(RbConfig) ? RbConfig : Config)::CONFIG['host_os'] =~ /mswin|win32|dos|mingw|cygwin/i
   WINDOWS = TRUE
+else
+  WINDOWS = FALSE
 end
 
 PREREQS = %w{}


melissa:
Ah shoot. Thank you!

melissa:
Soo, how do we get this promoted into PA? Does this warrant a new MCO release?

richardc:
Well I don't know much about how puppet-agent consumes things, but it looks like jenkins already promoted the merge commit as the version of mcollective: https://github.com/puppetlabs/puppet-agent/commit/36eca2f7458c571ab8281b615fbe12c01a77fc2e

When it comes to actually releasing, I think we'd need a tag as it's part of the repo.

melissa:
Yup, it's in PA#master at https://github.com/puppetlabs/puppet-agent/blob/master/configs/components/marionette-collective.json which is all I need. I'll let everyone else figure out if/when to release this. Thank you Richard!

PR #365, opened 2016-01-25
(MCO-748) Update mco.bat to reflect current standard

melissa:
With the switch over to building puppet-agent for windows with Vanagon,
we need to effectively install mcollective with install.rb. This script
installs mco.bat, which is a wrapper that lets windows handle ruby
scripts. This commit updates mco.bat to use our environment.bat file if
we install it, and to be more in line with the mco.bat file installed
with the puppet-agent package as built by puppet_for_the_win

This takes advantage of `environment.bat` in puppet-agent installed with https://github.com/puppetlabs/puppet-agent/pull/512

### DIFF
--- a/ext/windows/mco.bat
+++ b/ext/windows/mco.bat
@@ -1,7 +1,9 @@
 @echo off
-
 SETLOCAL
-
 call "%~dp0environment.bat" %0 %*
 
-%RUBY% -S -- mco %* --config "%CLIENT_CONFIG%"
+if [%CLIENT_CONFIG%] == [] (
+  ruby.exe -S -- mco %*
+) else (
+  ruby.exe -S -- mco %* --config "%CLIENT_CONFIG%"
+)
--- a/install.rb
+++ b/install.rb
@@ -280,7 +280,7 @@ def install_binfile(from, op_file, target)
 
   File.open(from) do |ip|
     File.open(tmp_file.path, "w") do |op|
-      op.puts "#!#{ruby}"
+      op.puts "#!#{ruby}" unless WINDOWS
       contents = ip.readlines
       contents.shift if contents[0] =~ /^#!/
       op.write contents.join


ext/windows/mco.bat
```diff
@@ -1,7 +1,8 @@
 @echo off
-
 SETLOCAL
-
-call "%~dp0environment.bat" %0 %*
-
-%RUBY% -S -- mco %* --config "%CLIENT_CONFIG%"
```
melissa: From what I could tell from the mco executable that's getting run here, the `--config` flag doesn't actually do anything... I didn't dig too deeply, so I could easily be wrong.

melissa:
still testing

ext/windows/mco.bat
richardc: It should be telling mco what configuration file to use, which looking at  https://github.com/puppetlabs/marionette-collective/blob/master/ext/windows/environment.bat#L5 is

```
SET CLIENT_CONFIG=%BASEDIR%\etc\client.cfg
```

That's also the default path that's set here:

https://github.com/puppetlabs/marionette-collective/blob/master/lib/mcollective/util.rb#L169-L171

Looking at the history it just points to issue 12555 (commit d127ab894) solving the problem of 'where should the config live' with an explicit path in mco.bat.  Later issue 20388 (5a83fb9708) solved it generally without undoing the explicit value in mco.bat to fully make use of the change.

melissa: Fun. So, I'm shaking things up, and we may no longer be consistent about how we're installing things on windows (https://github.com/puppetlabs/puppet-agent/blob/feature/master/windows-vanagon-development/configs/components/marionette-collective.rb#L98-L99). So, it's possible the new path would be ` C:\ProgramData\PuppetLabs\mcollective` rather than `C:\ProgramData\PuppetLabs\mcollective\etc`. I also don't think we'll be using the environment.bat file in the mco repo, but rather we'll have a global one available for all projects. Should we be setting the client and server config locations there?

(these path changes definitely aren't set in stone, they've just been proposed and potential ramifications are being discussed)

richardc: The current [path specs](https://github.com/puppetlabs/puppet-specifications/blob/master/file_paths.md) specify  `C:\ProgramData\PuppetLabs\mcollective\etc\client.cfg` too so I'd say that windows-vanagon branch is the odd one out here, not that change is impossible, but I would like to see it flow from specs into the implementation.

I'm +1 for dropping environment.bat if we can do that sanely.

melissa: The windows-vanagon branch definitely is the odd one out here! I'm not saying the path change is 100% going to happen, but it's on the table. I have a PR up at https://github.com/puppetlabs/puppet-specifications/pull/63 to try to get people talking about these changes and which ones are terrible and we absolutely can't do

ext/windows/mco.bat
```diff
@@ -1,7 +1,8 @@
 @echo off
-
 SETLOCAL
-
-call "%~dp0environment.bat" %0 %*
-
-%RUBY% -S -- mco %* --config "%CLIENT_CONFIG%"
+if exist "%~dp0environment.bat" (
+  call "%~dp0environment.bat" %0 %*
```
melissa: Will either use environment.bat from ext/windows in this repo or the one installed in the puppet-agent repo (https://github.com/puppetlabs/puppet-agent/pull/512)

ext/windows/mco.bat
melissa: I'll set this in the PA environment.bat https://github.com/puppetlabs/puppet-agent/pull/512/files#diff-f19be7ccad3d10336f31489de24d09ddR32

that'll change back to the current location, but I need to change how MCO is installed on windows

install.rb
```diff
@@ -280,7 +280,7 @@ def install_binfile(from, op_file, target)
 
   File.open(from) do |ip|
     File.open(tmp_file.path, "w") do |op|
-      op.puts "#!#{ruby}"
+      op.puts "#!#{ruby}" unless WINDOWS
```
melissa: Turns out, we don't want this line in `mco.bat`.

melissa:
cc/ @MikaelSmith @kylog @joshcooper could one of you review this?

melissa:
ping @richardc @MikaelSmith @joshcooper and who ever else might be able to review this!

ext/windows/mco.bat
```diff
@@ -1,7 +1,8 @@
...
-
-call "%~dp0environment.bat" %0 %*
-
-%RUBY% -S -- mco %* --config "%CLIENT_CONFIG%"
+if exist "%~dp0environment.bat" (
+  call "%~dp0environment.bat" %0 %*
+) else (
+  SET PATH=%~dp0;%PATH%
+)
+ruby.exe -S -- mco %* --config "%CLIENT_CONFIG%"
```
richardc: Isn't this going to break in the cases where there is no environment.bat to provide %CLIENT_CONFIG%?

melissa: Oh, I hadn't thought of that. I suppose so. Should we assume there will always be an environment.bat? You provide it in the mco repo and we provide it in the puppet-agent repo. Should I take out the check for it? Or should I add a check to see if `%CLIENT_CONFIG% is set?

richardc: I'd say that if we're always providing an environment.bat then we should remove the conditional load of it from lines 3-6. 

We could check that CLIENT_CONFIG is set too, as if the environment.bat that is loaded lacks the value then we'll have an invalid command line, but maybe we can just not use the --config option at all as it's duplicating the default.

I think it opens the wider can of worms - is environment.bat intended as configuration - would a user have a solid use-case about editing it and setting CLIENT_CONFIG to point elsewhere?

melissa: Updated to check if `%CLIENT_CONFIG%` is set.

As to `environment.bat` used as configuration, I have no idea. It's not going to be preserved on upgrade, so any changes a user makes to it will be overwritten. Though if we remove that capacity, anyone currently using it as such (going off the `environment.bat` file in this repo) might get mad at us if we remove that ability? Puppet_for_the_win currently does not set this at all. Maybe keep this functionality but don't endorse `environment.bat` as something that can be used as a configuration element? Remove any configuration settings from https://github.com/puppetlabs/puppet-agent/pull/512/files but still let users set them as defined in https://github.com/puppetlabs/marionette-collective/blob/master/ext/windows/environment.bat ?

ext/windows/mco.bat
```diff
@@ -1,7 +1,11 @@
...
 call "%~dp0environment.bat" %0 %*
 
-%RUBY% -S -- mco %* --config "%CLIENT_CONFIG%"
+if [%CLIENT_CONFIG%] == [] (
+  set CONFIG="--config %CLIENT_CONFIG%"
+) else (
+  set CONFIG=""
+)
+
+ruby.exe -S -- mco %* "%CONFIG%"
```
richardc: I think this is going to go terribly when CLIENT_CONFIG has spaces in it, and also we're now supplying a blank arg in both cases.

As ugly as it looks, I think this is the safer way to write this:

```
if [%CLIENT_CONFIG%] == [] (
 ruby.exe -S -- mco %* --config "%CLIENT_CONFIG%"
) else (
  ruby.exe -S -- mco %*
)
```

Though now I've written it out that way, the condition looks like it's backwards.

ext/windows/mco.bat
richardc: I was more thinking along the lines of "drop CLIENT_CONFIG from environment.bat".  With the current codebase it's just a restating of the in-code default, which has the possibility to confuse.  If users aren't using environment.bat as config, then there's no harm in dropping the setting, or the file completely.

melissa:
@richardc updated!

richardc:
@melissa Merged!

melissa:
WOO! Thanks!

PR #366, opened 2016-02-16
(MCO-751) Manage PID files to avoid race conditions on process start

underscorgan:
The daemonize_runner was not using an exclusive create when writing its pid
file so you could get into a race condition with SysVInit by running
`service mcollective start & service mcollective start`. This also adds
checking to make sure stale PID files get cleaned up.

### DIFF
--- a/lib/mcollective/unix_daemon.rb
+++ b/lib/mcollective/unix_daemon.rb
@@ -19,10 +19,40 @@ module MCollective
 
       UnixDaemon.daemonize do
         if pid
-          begin
-            File.open(pid, 'w') {|f| f.write(Process.pid) }
-          rescue Exception => e
+          # Clean up stale pidfile if needed
+          if File.exist?(pid)
+            lock_pid = File.read(pid)
+            begin
+              lock_pid = Integer(lock_pid)
+            rescue ArgumentError, TypeError
+              lock_pid = nil
+            end
+
+            # If there's no pid in the pidfile, remove it
+            if lock_pid.nil?
+              File.unlink(pid)
+            else
+              begin
+                # This will raise an error if the process doesn't
+                # exist, and do nothing otherwise
+                Process.kill(0, lock_pid)
+                # If we reach this point then the process is running.
+                # We should raise an error rather than continuing on
+                # trying to create the PID
+                raise "Process is already running with PID #{lock_pid}"
+              rescue Errno::ESRCH
+                # Errno::ESRCH = no such process
+                # PID in pidfile doesn't exist, remove pidfile
+                File.unlink(pid)
+              end
+            end
+
           end
+
+          # Use exclusive create on the PID to avoid race condition
+          # when two mcollectived processes start at the same time
+          opt =  File::CREAT | File::EXCL | File::WRONLY
+          File.open(pid, opt) {|f| f.print(Process.pid) }
         end
 
         begin
--- a/spec/unit/mcollective/unix_daemon_spec.rb
+++ b/spec/unit/mcollective/unix_daemon_spec.rb
@@ -13,9 +13,9 @@ module MCollective
 
       it "should write the pid file if requested", :unless => MCollective::Util.windows? do
         f = mock
-        f.expects(:write).with(Process.pid)
+        f.expects(:print).with(Process.pid)
 
-        File.expects(:open).with("/nonexisting", "w").yields(f)
+        File.expects(:open).with("/nonexisting", File::CREAT | File::EXCL | File::WRONLY).yields(f)
 
         r = mock
         r.expects(:main_loop)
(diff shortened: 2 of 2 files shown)


lib/mcollective/unix_daemon.rb
```diff
@@ -19,10 +19,35 @@ def self.daemonize_runner(pid=nil)
...
+                # Errno::ESRCH = no such process
+                # PID in pidfile doesn't exist, remove pidfile
+                File.unlink(pid)
+              end
+            end
           end
+
+          # Use exclusive write on the PID to aviod race condition
+          # when two processes start at the same time
+          opt =  File::CREAT | File::EXCL | File::WRONLY
```
melissa: Would we want to add `LOCK_EX` to this? I feel like we'd have to unlock it afterward though, and I'm not sure how worth it that'd be http://ruby-doc.org/core-2.1.0/File.html#method-i-flock

lib/mcollective/unix_daemon.rb
```diff
@@ -19,10 +19,35 @@ def self.daemonize_runner(pid=nil)
...
+
+            # If there's no pid in the pidfile, remove it
+            if lock_pid.nil?
+              File.unlink(pid)
+            else
+              begin
+                # This will raise an error if the process doesn't
+                # exist, and do nothing otherwise
+                Process.kill(0, lock_pid)
+              rescue Errno::ESRCH
```
melissa: According to http://ruby-doc.org/core-2.1.0/Process.html#method-c-kill, `RangeError` could also indicate an invalid pid, though not exclusively. I'm not sure if it would be worth adding `RangeError` to this rescue or not

melissa:
This seems to undermine existing expectations in the unit tests :(

lib/mcollective/unix_daemon.rb
underscorgan: Yeah, I'm not sure how worth it it would be to need to deal with unlocking. I was mostly trying to be consistent with what they were doing in Puppet (https://github.com/puppetlabs/puppet/blob/c384ddacdc24167f6395cc4c29e1dd28a7ab4018/lib/puppet/file_system/file_impl.rb#L41-L44)

lib/mcollective/unix_daemon.rb
underscorgan: Hm, not sure for this. Was trying to be consistent with Puppet again https://github.com/puppetlabs/puppet/blob/d2199af906fa5ca19ec0c37c82794934ea4fb4b5/lib/puppet/util/pidlock.rb#L49

underscorgan:
Working on test updates and some new tests now.

lib/mcollective/unix_daemon.rb
melissa: Cool, fair enough

lib/mcollective/unix_daemon.rb
melissa: Fair enough

lib/mcollective/unix_daemon.rb
```diff
@@ -19,10 +19,42 @@ def self.daemonize_runner(pid=nil)
...
+              File.unlink(pid)
+            else
+              begin
+                # This will raise an error if the process doesn't
+                # exist, and do nothing otherwise
+                Process.kill(0, lock_pid)
+                # If no error is raised here the pid is not stale
+                # We should raise an error rather than continuing on
+                # trying to create the PID
+                locked_and_not_mine = true
```
melissa: So, if we successfully kill the service, we shouldn't try to start it up again?

underscorgan: Killing it with a signal of 0 will just do error checking, so the process won't actually get killed.

melissa: Oooh, that's what I was missing! Cool, then this makes a lot of sense.

melissa:
:+1: pending travis, this makes a lot of sense to me.

lib/mcollective/unix_daemon.rb
```diff
@@ -19,10 +19,42 @@ def self.daemonize_runner(pid=nil)
 
       UnixDaemon.daemonize do
         if pid
-          begin
-            File.open(pid, 'w') {|f| f.write(Process.pid) }
-          rescue Exception => e
+          # Clean up stale pidfile if needed
+          if File.exist?(pid)
+            lock_pid = File.read(pid)
+            locked_and_not_mine = false
```
richardc: This is a poor choice of name.  We're not really checking if anything is locked, we're checking if the pidfile has a pid in it.  Also because we're reading it to check we can be pretty sure it's never going to be 'mine'.  Something more like `pid_present_and_running` (or eliminating it completely, see note to follow) would be more apt.

lib/mcollective/unix_daemon.rb
```diff
@@ -19,10 +19,42 @@ def self.daemonize_runner(pid=nil)
...
+                # trying to create the PID
+                locked_and_not_mine = true
+              rescue Errno::ESRCH
+                # Errno::ESRCH = no such process
+                # PID in pidfile doesn't exist, remove pidfile
+                File.unlink(pid)
+              end
+            end
+
+            raise "Could not create PID file: #{pid}" if locked_and_not_mine
```
richardc: So since this can only become true in one way/place, wouldn't it be better to move this up to line 43, and eliminate the poorly named variable?

lib/mcollective/unix_daemon.rb
```diff
@@ -19,10 +19,42 @@ def self.daemonize_runner(pid=nil)
...
           end
+
+          # Use exclusive write on the PID to aviod race condition
+          # when two processes start at the same time
```
richardc: Typo - `aviod` for `avoid`.

lib/mcollective/unix_daemon.rb
richardc: This isn't actually locking anything, it's just an exclusive create.  It means if there is a race and and some other process creates the pidfile that the second process will fail so it's sufficient

From [open(2)](http://linux.die.net/man/2/open)
> O_EXCL
> Ensure that this call creates the file: if this flag is specified in conjunction with O_CREAT, and pathname already exists, then open() will fail.

richardc:
Looks OK, but it's not quite correct to call this an exclusive lock in the commit message.  This technique is an exclusive create.

underscorgan:
Hey @richardc , I've addressed your comments/nits and updated the commit message, so let me know how this looks!

lib/mcollective/unix_daemon.rb
```diff
@@ -19,10 +19,40 @@ def self.daemonize_runner(pid=nil)
...
+              File.unlink(pid)
+            else
+              begin
+                # This will raise an error if the process doesn't
+                # exist, and do nothing otherwise
+                Process.kill(0, lock_pid)
+                # If no error is raised here the pid is not stale
+                # We should raise an error rather than continuing on
+                # trying to create the PID
+                raise "Could not create PID file: #{pid}"
```
richardc: Is this an accurate message?   We're not failing to create a pid file, we're deliberately raising because the existing pid file refers to a process we're allowed to deliver a signal to.

lib/mcollective/unix_daemon.rb
```diff
@@ -19,10 +19,40 @@ def self.daemonize_runner(pid=nil)
...
+
+            # If there's no pid in the pidfile, remove it
+            if lock_pid.nil?
+              File.unlink(pid)
+            else
+              begin
+                # This will raise an error if the process doesn't
+                # exist, and do nothing otherwise
+                Process.kill(0, lock_pid)
+                # If no error is raised here the pid is not stale
```
richardc: The double negative here hurts my addled brain.  May I suggest

`If we reach this point then the process is running.`

underscorgan:
@richardc ok, updated again!

richardc:
@underscorgan cool, thanks.

PR #367, opened 2016-02-17
(MCO-634) Set build defaults to LTS targets only

melissa:
We are no longer using the packaging repo to build mco packages. This
happens as a part of the puppet agent build process. This means we no
longer use the build target definitions stored in ext/build-defaults to
determine what platforms to build mco packages for. However, this is a
tool available for community use. Rather than remove these entries
entirely, we should change them to be only LTS platforms. That way, if a
community member is using this automation, they can easily edit the file
to build the platforms they want. It also helps us when we inevitably
look back at this file and wonder why we have build targets listed that
went EOL many cycles ago.


melissa:
Totally not a priority to get this in

PR #370, opened 2016-02-25
(MCO-753) The init scripts should not assume the existence of the pid dir

shrug:
Currently, the init scripts naively assume that the pid directory
exists. This works well enough, assuming puppet starts first after
a reboot and creates the pid dir. This is not always the case.

This updates the init scripts to create the pid directory
if it does not exist.

### DIFF
--- a/ext/aio/debian/mcollective.init
+++ b/ext/aio/debian/mcollective.init
@@ -24,7 +24,8 @@ uid=`id -u`
 
 
 # PID directory
-pidfile="/var/run/puppetlabs/mcollectived.pid"
+piddir="/var/run/puppetlabs"
+pidfile="${piddir}/mcollectived.pid"
 
 name="mcollective"
 mcollectived=/opt/puppetlabs/puppet/bin/mcollectived
@@ -52,6 +53,7 @@ fi
 case "$1" in
     start)
         echo "Starting daemon: " $name
+        mkdir -p ${piddir}
         # start the program
         start-stop-daemon --start --pidfile ${pidfile} --oknodo --quiet --startas ${mcollectived} -- ${daemonopts} --daemonize
         [ $? = 0 ] && { exit 0 ; } || { exit 1 ; }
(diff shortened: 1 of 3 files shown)


underscorgan:
:+1:

ext/aio/suse/mcollective.init
```diff
@@ -32,7 +32,8 @@
 desc=${DESC:-mcollective daemon}
 daemon=${DAEMON:-/opt/puppetlabs/puppet/bin/mcollectived}
 name=${NAME:-mcollectived}
-pidfile=${PIDFILE:-/var/run/puppetlabs/mcollectived.pid}
+piddir=${PIDDIR:-/var/run/puppetlabs/}
```
kylog: Do you need the trailing `/` in this one?

kylog:
Jira-fu: should this be moved to MCO?

shrug:
It should, yes, and it's now MCO-753

kylog:
LGTM. Will merge once travis does its thing.

kylog:
We should also do the docs needful a la https://github.com/puppetlabs/marionette-collective/pull/369

parisiale:
@kylog I can do that

ext/aio/debian/mcollective.init
```diff
@@ -52,6 +53,7 @@ fi
 case "$1" in
     start)
         echo "Starting daemon: " $name
+        mkdir -p ${piddir}
```
mckern: I was going to say something about unquoted variables but it feels like that's trying to boil the ocean in this case.

flaf:
Hi, indeed it seems to me that `mkdir -p "$piddir"` could be more sure but it's probably a detail.

Furthermore, I have added a comment in the ticket https://tickets.puppetlabs.com/browse/MCO-753?#comment-268494. It seems to me that the content of the file `/etc/default/mcollective` is still irrelevant because the  `START` variable and the `DAEMON_OPTS` are completely ignored by the init script.

Regards.

PR #372, opened 2016-03-31
SSL security: use SSH agent when available

raphink:
When using the SSL security plugin, it is useful to use SSH keys (because pretty much everyone has one) as the SSL keypair for users.

This patch makes `mco` look for the key in ssh-agent to avoid entering the key password every time.


raphink:
@kylog any thoughts on that?

MikaelSmith:
I'd appreciate a ticket filed around this. Also can you rebase to pick up fixes for Travis CI?

It seems useful, but I'd like a 2nd review. ping @adrienthebo

lib/mcollective/security/ssl.rb
```diff
@@ -306,7 +307,29 @@ def client_cert_dir
...
+          a = Net::SSH::Authentication::Agent.connect
+          # Look for key based on public key
+          pub_key = OpenSSL::PKey::RSA.new File.read client_public_key
+          a.identities.each do |i|
+            if pub_key.fingerprint == i.public_key.fingerprint
+              sig =  a.sign(i, body.to_s)
+              return SSL.base64_encode(parse_ssh_sig(sig))
+            end
+          end
+        rescue
```
adrienthebo: Can we rescue a more specific exception here? I don't want us accidentally rescuing an ArgumentError and treating that as a failure to connect to the SSH agent.

lib/mcollective/security/ssl.rb
```diff
@@ -306,7 +307,29 @@ def client_cert_dir
       def makehash(body)
         Log.debug("Creating message hash using #{private_key_file}")
 
-        sign(private_key_file, body.to_s)
+        begin
+          a = Net::SSH::Authentication::Agent.connect
+          # Look for key based on public key
+          pub_key = OpenSSL::PKey::RSA.new File.read client_public_key
+          a.identities.each do |i|
```
adrienthebo: Variable names like `i` and `a` increase the cognitive load on any readers. I personally tripped over `pub_key.fingerprint == i.public_key.fingerprint` because of this abbreviation. This is just a small style nit but I think it would be helpful for following readers if we expanded the variable names.

lib/mcollective/security/ssl.rb
```diff
@@ -306,7 +307,29 @@ def client_cert_dir
...
+              sig =  a.sign(i, body.to_s)
+              return SSL.base64_encode(parse_ssh_sig(sig))
+            end
+          end
+        rescue
+          sign(private_key_file, body.to_s)
+        end
+      end
+
+      def parse_ssh_sig(string)
```
adrienthebo: I honestly have no idea what this function is doing. What value are we parsing from the SSH signature?

raphink: Unfortunately, SSH does not use standard SSL formats. Iirc (I coded this a months and a half ago), in this case, the signature returned by SSH is a binary format that needs to be parsed and reformatted in order to be shown as OpenSSL RSA pub keys.

adrienthebo: Ah, that makes a ton of sense. (Isn't it great how SSH does fun things like that?) Given that this code is doing some low level string processing this sort of context could be lost very easily; could we add this as a comment on the method for future readers?

joshcooper: Does mcollective support DSA keys? If so, I don't think the `parse_ssh_sig` takes that into account. The [ssh protocol docs](http://cvsweb.openbsd.org/cgi-bin/cvsweb/src/usr.bin/ssh/PROTOCOL.agent?rev=1.10&content-type=text/x-cvsweb-markup) define an opaque `signature_blob`:

```
The signature will be returned in the response message:

	byte			SSH2_AGENT_SIGN_RESPONSE
	string			signature_blob

If the agent cannot find the key specified by the supplied key_blob then
it will return SSH_AGENT_FAILURE.
```

whose format is described in https://tools.ietf.org/html/rfc4253#page-13. For example, ssh-dsa signatures are defined to be:

```
   The resulting signature is encoded as follows:

      string    "ssh-dss"
      string    dss_signature_blob

 The value for 'dss_signature_blob' is encoded as a string containing
   r, followed by s (which are 160-bit integers, without lengths or
   padding, unsigned, and in network byte order).
```

Is there a higher-level API we can use to extract the signature, like `EVP_` for openssl?

lib/mcollective/security/ssl.rb
```diff
@@ -306,7 +307,29 @@ def client_cert_dir
...
+        offset = 0
+        typelen = string[offset..(offset+3)].reverse.unpack("L")[0]
```
joshcooper: I think the `reverse` assumes the host is little-endian, so that wouldn't work on big-endian architectures like powerpc (not sure if we support mco there?).

MikaelSmith: AIX, so I think so.

mruzicka: using 'N' as the unpack format would make the expression portable:
```ruby
        typelen = string[offset..(offset+3)].unpack("N")[0]
```

lib/mcollective/security/ssl.rb
```diff
@@ -306,7 +307,29 @@ def client_cert_dir
...
+        offset += 4
+        offset += typelen
+        siglen = string[offset..(offset+3)].reverse.unpack("L")[0]
```
mruzicka: again: using `N` as the unpack format makes the epxression portable

mruzicka:
@raphink Could you please rebase the PR, and use the ```unpack("N")``` as per my comment above?

raphink:
@mruzicka done

lib/mcollective/security/ssl.rb
```diff
@@ -1,5 +1,6 @@
 require 'base64'
 require 'openssl'
+require 'net/ssh'
```
MikaelSmith: Does this run? When I try to do `bundle console` and `require 'net/ssh'`, I get a failure to find the library.

raphink:
I'll have a look

bastelfreak:
ping @raphink. Can you take a look at this again?

lib/mcollective/security/ssl.rb
raphink: Should I add `net-ssh` in the Gemfile then?

lib/mcollective/security/ssl.rb
raphink: Yes, I'd just have to find out which exception to rescue.

lib/mcollective/security/ssl.rb
raphink: OK

lib/mcollective/security/ssl.rb
raphink: I haven't found such an API.

lib/mcollective/security/ssl.rb
raphink: Changed the variable names.

raphink:
Just rebased the branch

lib/mcollective/security/ssl.rb
MikaelSmith: There's some work needed in puppet-agent to include it as well. We could also make it optional - i.e. rescue the require and disable this functionality if not available.

raphink:
@bastelfreak did you take a look since I rebased?

PR #374, opened 2016-05-02
(MCO-736) Singletarget

reguero:
Implements the suggestion in MCO-736 to use a single target for all agents in the activemq connector.
We are running a collective of 6165 nodes with this change since a few weeks.
We observe that the number of subscriptions indeed dropped from ~75k to ~20k. 
The brokers are happier and querying the list of subscriptions via JMX is now faster.

```
                                      Best regards    ...Ignacio...
```

### DIFF
--- a/website/reference/plugins/connector_rabbitmq.md
+++ b/website/reference/plugins/connector_rabbitmq.md
@@ -333,3 +333,11 @@ The SSL ciphers to use when communicating with the middleware.
 
 - _Default:_ no default
 - _Allowed values:_ A string supplying an [OpenSSL cipher suite][cipherstrings]
+
+#### `plugin.rabbitmq.agents_multiplex`
+
+Whether to use a single target for all agents in a node thus reducing the number of subscriptions in the message broker
+
+- _Default:_ false
+- _Allowed values:_ A boolean value
+
--- a/website/reference/plugins/connector_rabbitmq.md
+++ b/website/reference/plugins/connector_rabbitmq.md
@@ -336,7 +336,8 @@ The SSL ciphers to use when communicating with the middleware.
 
 #### `plugin.rabbitmq.agents_multiplex`
 
-Whether to use a single target for all agents in a node thus reducing the number of subscriptions in the message broker
+Whether to use a single target for all agents in a node. This is an optimization that may make sense when running collectives with several thousands of nodes in order to reduce the number of subscriptions in the message broker.
+This is a trade-off between increasing network traffic by delivering messages to all nodes - and letting them select messages they care about - versus increasing work in the message broker to handle large numbers of subscriptions.
 
 - _Default:_ false
 - _Allowed values:_ A boolean value
--- a/lib/mcollective/connector/activemq.rb
+++ b/lib/mcollective/connector/activemq.rb
@@ -541,6 +541,7 @@ module MCollective
         raise("Unknown target type #{type}") unless [:directed, :broadcast, :reply, :request, :direct_request].include?(type)
         raise("Unknown collective '#{collective}' known collectives are '#{@config.collectives.join ', '}'") unless @config.collectives.include?(collective)
 
+        agents_multiplex = get_bool_option("activemq.agents_multiplex", "false")
         target = {:name => nil, :headers => {}}
 
         case type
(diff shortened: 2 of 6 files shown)


MikaelSmith:
Thanks for making a go at this! Would you mind rebasing? The CI failures have been fixed on #master. Also, please take a look at the CLA; did you change your user name?

reguero:
I did the rebase. Also the IT department leader just told me that he signed the CLA for CERN last week so it is OK for me.  I have just mentioned this to the CLA Submissions Team.

MikaelSmith:
Thanks. I'm planning to do a little more investigating to make sure I understand this before I merge, hopefully get to that this week.

reguero:
FYI, in case it helps, to check how this works in the server code, I saw that the receiver_thread of the runner creates the agents with Agents.new.

The initialize method of class Agents calls load agents that calls loadagent(agentname) for each agent that it finds in "#{libdir}/mcollective/agent".
loadagent(agentname) calls once Util.subscribe(Util.make_subscriptions(agentname, :broadcast)).

Util.make_subscriptions(agentname, :broadcast) will give a list of hashes as {:agent => agent, :type => :broadcast, :collective => c} for all collectives.

Util.subscribe will call connection.subscribe(target[:agent], target[:type], target[:collective]) for each of these.

subscribe as provided by the activemq connector will call make_target(agent, type, collective) to define the source and then @connection.subscribe(source[:name], source[:headers], source[:id]). As we modified make_target, when using plugin.activemq.agents_multiplex = true the source[:name] contains the shared target.
Please note that the code only calls @connection.subscribe once per source[:id] that in our case is equal to the source[:name] so although there will be a call to subscribe for each agent, the subscription will only be created the first time.

All this is compatible with the receiver_thread of the runner calling the receive method that calls @connection.receive as receive as provided by the activemq connector will not check for which target the message came. The receiver_thread will call afterwards agentmsg(request) on the request that we received which will call @agents.dispatch(request, @connection) to dispatch it to the right agent and handle the reply if any with reply(reply_message, request).

MikaelSmith:
Could you add a section to the documentation at https://github.com/puppetlabs/marionette-collective/blob/master/website/reference/plugins/connector_activemq.md on this new parameter?

It looks like a similar change could be made to the RabbitMQ connector. It'd be nice to have parity in them.

reguero:
I added a commit with similar code for rabbitmq as well as the doc entry. 
Please bear in mind that do no run rabbitmq so I have no means to test it in production like that activemq one.

reguero:
Hi @MikaelSmith and all,
 No plan to merge this one?

MikaelSmith:
I'm making time for it this week.

MikaelSmith:
@ripienaar you weighed in on the ticket, have any thoughts on this PR? It doesn't break anything, and seems like a reasonable way to customize dealing with large collectives.

@reguero is the main impact of your approach to handling collectives to speed up querying the list? Or are there other benefits as well?

ripienaar:
It's fine but should be a default off option (as it is in the code so 👍), the issues is lets say you have  the r10k agent, it will be on a tiny fraction of hosts.  

The per agent topics are there so that broadcasts to this agent only goes to those nodes and not all over the show, so its a network traffic load optimisation.

But all optimisations have tradeoffs and times they make sense, for sure if you have 60k topic subs as a result to your size and how ActiveMQ uses memory it will be better to optimise the subs count rather than network broadcasts.

So the option is good, we should be sure to call out the trade off in the docs and I think the current behaviour is best default for the common case.

ripienaar:
So to be clear: I think we need better doc updates that call out the when/why of this option, on the surface to someone skimming the docs - `Whether to use a single target for all agents in a node thus reducing the number of subscriptions in the message broker.` sounds like its always a good idea, but it's a good idea on the top end of the scale 

To clarify the broadcast to just one agent topic comment - since 2.0ish we have direct addressing that sends messages directly to nodes into their queues by default, but once it discovers I think > 10 nodes it will broadcast to them over these topics as a optimisation over sticking 1000s of messages in queues.

website/reference/plugins/connector_rabbitmq.md
```diff
@@ -333,3 +333,11 @@ The SSL ciphers to use when communicating with the middleware.
 
 - _Default:_ no default
 - _Allowed values:_ A string supplying an [OpenSSL cipher suite][cipherstrings]
+
+#### `plugin.rabbitmq.agents_multiplex`
+
+Whether to use a single target for all agents in a node thus reducing the number of subscriptions in the message broker
```
MikaelSmith: Expand it by adding something like the following?

`This is a trade-off between increasing network traffic by delivering messages to all nodes - and letting them select messages they care about - versus increasing work in the message broker to handle large numbers of subscriptions.`

reguero: Hi @MikaelSmith 
Thanks for the suggestion,
I also thought to update the doc line before in the sense of the comments by @ripienaar 
Would the text below be OK for you?

'Whether to use a single target for all agents in a node. This is an optimization that may make sense when running collectives with several thousands of nodes in order to reduce the number of subscriptions in the message broker.
This is a trade-off between increasing network traffic by delivering messages to all nodes - and letting them select messages they care about - versus increasing work in the message broker to handle large numbers of subscriptions.'

MikaelSmith: Sounds good to me. If that's an accurate description of your experience with it, then 👍

reguero:
Updated the doc as discussed and rebased

MikaelSmith:
Thanks @reguero! Very much appreciated.

PR #378, opened 2016-05-10
(maint) fix activemq install

er0ck:
Before this change, the pre-suite would not install activemq correctly, nor tell you why (version out of date, in this case).  

This change bumps the version, and ensures curl will fail when receiving >400 from the apache servers when fetching activemq.
[skip ci]


MikaelSmith:
👍

MikaelSmith:
Is this needed for tests on stable?

er0ck:
i don't think master nor stable run anything at all in our jenkins jobs 😒 
(oh, they do a bundle install 👍 💯 ✅ )

maybe they could at least run the pre-suite?  
i just fixed this as i was doing some validation and it made it much easier to just be able to run the pre-suite to install/setup

PR #380, opened 2016-05-19
(MCO-763) Move prepare_installation to run first

McdonaldSeanp:
prepare_installation does not depend on any variable definitions before it, so
it needs to be moved to the first action taken in install.rb to avoid
situations where the script is using install options before they are defined


mruzicka:
LGTM

PR #386, opened 2016-06-24
(RE-7302) Update Stomp gem dependency to >= 1.4.1

shrug:
The Stomp 1.4.1 gem adds fixes for SSL hostname verification


underscorgan:
👍

PR #387, opened 2016-07-27
Updates to make interbroker links more stable.

dforste:
Updates to messageEvictionStrategy to prevent buildup of messages in queue for topics on larger clusters.

### DIFF
--- a/ext/activemq/examples/multi-broker/broker1-activemq.xml
+++ b/ext/activemq/examples/multi-broker/broker1-activemq.xml
@@ -34,13 +34,31 @@
 
           For more information, see:
           http://activemq.apache.org/producer-flow-control.html
+
+          Additional hints from this guy running a much larger cluster than me.
+          http://grokbase.com/t/gg/mcollective-users/163xya3hgq/activemq-topology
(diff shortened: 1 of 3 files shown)


dforste:
This relates to https://tickets.puppetlabs.com/browse/MCO-58

dforste:
This seems to fall under the Trivial Patch Exemption Policy as it "introduce no new functionality."

MikaelSmith:
It should be fine; submitting commits via the web interface doesn't work well with our bot either.

dforste:
Ya not a problem the biggest part is the addition of a pruning strategy otherwise topics keep an ungodly amount of messages around. This seems to be the cause of some instability. Another recommendation is to upgrade to a full mesh as that helps with intermittent network or link issues. 

````
<messageEvictionStrategy>
  <oldestMessageEvictionStrategy/>
</messageEvictionStrategy>
<pendingMessageLimitStrategy>
  <prefetchRatePendingMessageLimitStrategy
    multiplier="2"
  />
</pendingMessageLimitStrategy>
````

PR #390, opened 2016-08-17
(maint) Update Ruby testing to currently supported versions

MikaelSmith:
2.0.0 was end-of-lifed in February, so support latest patches of all
later versions.


MikaelSmith:
@kylog got some issues to fix :)

McdonaldSeanp:
@MikaelSmith this works great (I also tested the specs locally and everything is good) 👍

PR #391, opened 2016-08-23
(MCO-773) Add mcollective.gemspec so we can generate a gem

tampakrap:


MikaelSmith:
@McdonaldSeanp does this make sense? We already generate a gem, how does that work?

McdonaldSeanp:
Yeah, we dont generate an MCO gem as of now

ripienaar:
There has never been a mcollective gem only a client gem. This is on purpose because you can't use such a thing - mcollective is a daemon with confit requirements, needs a service script and more. You can't deliver a working mcollective as a gem

mruzicka:
@tampakrap why do you need the gem, and how do you address the issues mentioned by @ripienaar?

tampakrap:
take a look at the discussion at the ticket please. I'll be glad to provide further info as well https://tickets.puppetlabs.com/browse/MCO-773

PR #397, opened 2016-09-19
(MCO-775) Include discovery timeout when making requests.

adreyer:
The discovery timeout sets a base level for how long it takes agents to
respone to requests. It should be included along with any filters when
computing the discovrey timeout for all requests. Before this commit
only the --timeout or the timeout of specific agents was used. Now the
discovery timeout is added to that time to account for system wide
performance.

### DIFF
--- a/lib/mcollective/client.rb
+++ b/lib/mcollective/client.rb
@@ -168,7 +168,10 @@ module MCollective
 
       @options = options if options
       threaded = @options[:threaded]
-      timeout = @discoverer.discovery_timeout(@options[:timeout], @options[:filter])
+      # The discovery timeout should be added to the timeout to allow global
+      # configuration in environments when nodes are slow to respond
+      dtimeout = @options[:disctimeout] || @config.discovery_timeout || 0
+      timeout = @discoverer.discovery_timeout(@options[:timeout] + dtimeout, @options[:filter])
       request = createreq(body, agent, @options[:filter])
       publish_timeout = @options[:publish_timeout]
       stat = {:starttime => Time.now.to_f, :discoverytime => 0, :blocktime => 0, :totaltime => 0}


MikaelSmith:
This seems reasonable to me... this means the semantics are:
- discovery_timeout: time we expect it to take for all recipients to respond (and for mcollective to process those responses)
- timeout: a way to override how long a module is expected to run on an agent (modules have their own configured timeouts)

Combining those two gives us the time we should wait for the client request to complete (which is what your change does).

ripienaar:
Curious about actual commands to reproduce this, from what I know there should be very few cases actually hitting this since the RPC level requests do consider these I think

MikaelSmith:
Deployments over WAN connections seem to need to bump up the `--timeout` flag or they only get a partial set of responses. One report was doing `mco service ...` with ~4000 nodes, and they'd see less than half respond.

It actually may be an issue with the default timeout for `mco service`, as it mainly seemed to be Windows nodes didn't respond quickly enough.

ripienaar:
Typically for most commands - and certainly mco service - the timeout option will be passed into the client from the RPC::Client class with the logic found at https://github.com/puppetlabs/marionette-collective/blob/master/lib/mcollective/rpc/client.rb#L87-L104

So I think this PR will in effect do this twice.

ripienaar:
And that code might have a bug or omission but it's probably there where the problem lies with `mco service` related stuff.  It does not seem to me this fix will actually fix the described problem since the linked code already exist to deal with this hypothesis

adreyer:
The was also a problem with  "mco ping" which doesn't seem to use rpc::client

ripienaar:
Yes this fix will fix mco ping. Probably the best fix is to remove the special handling from RPC::Client I linked to and putting this here if you checked and see doing it here will address all the cases the RPC::Client code handles? That way we don't dupe the fix

ripienaar:
From what I recall the reason this logic is in the RPC client only and not in the M::Client is that we really do need to do what the user asks. If dtimeout is let's say 10 and someone does --timeout 10 he would reasonably expect the program to do what it's told in the case of mco ping for e.g. instead of a 20 second timeout. This is super important I think so would be reluctant to do something to change that. 

RPC clients are a bit more magical so tweaks there are ok but still the CLI wins - if given a timeout then listen to the user and I think that's right. 

Normal RPC lifecycle do consider this additional network latency so that's fine too from my POV - assuming there is no bug in how RPC::Client calculates that.

adreyer:
I created a separate PR to just fix ping.

https://github.com/puppetlabs/marionette-collective/pull/398

adreyer:
closed in favor of just fixing ping

PR #398, opened 2016-09-20
(MCO-775) Ping application should include discovery_timeout

adreyer:
This is a much more limited alternative to https://github.com/puppetlabs/marionette-collective/pull/397

### DIFF
--- a/lib/mcollective/application/ping.rb
+++ b/lib/mcollective/application/ping.rb
@@ -49,6 +49,11 @@ module MCollective
     end
 
     def main
+      # If the user did not override the default timeout include the discovery timeout
+      if options[:timeout] == 5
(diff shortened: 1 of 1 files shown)


ripienaar:
still seems to break the #1 rule for me - do what the user says.  If he gives --timeout 10, it should take 10 seconds.  If not, sure mess around and adjust it to what make sense given the environment.

So I'd look at limiting this to cases where the user did not specify a timeout at all. There's no proper way to figure this out but when not given it would be `5` in the options, thats what RPC::Client also uses for this

ripienaar:
👍

adreyer:
I added a new commit to only include discovery timeout if the timeout is the default. If that seems better I'll squash it down.

PR #399, opened 2016-10-11
(maint) Add comment about updating limits with systemd

underscorgan:


MikaelSmith:
Seems useful. I suspect there are other places this should apply, but I don't see any off the top.

MikaelSmith:
I think I'd like a ticket for this, to track the conceptual idea (documenting updating limits with systemd) in case we later realize it's needed elsewhere.

underscorgan:
@MikaelSmith there's an internal RE ticket for this, and I've updated it in puppet, mco, pxp-agent, and ezbake. I didn't want to reference an internal ticket, and didn't know if it was useful to open tickets for everything ¯\_(ツ)_/¯

MikaelSmith:
Ok

underscorgan:
woo thanks!

PR #400, opened 2016-10-21
(MCO-776) Support PQL in STDIN discovery

ripienaar:
This enables using Puppet Query output as discovery source for
mcollective

```
puppet query "inventory { facts.os.name = 'CentOS' }"| mco puppet runonce
```

Results from puppet query has to include a certname items.

This also makes the find application STDIN aware and adds a nice
reusable method to do the switch other apps can use who wish to support
this workflow

### DIFF
--- a/lib/mcollective/application/find.rb
+++ b/lib/mcollective/application/find.rb
@@ -6,6 +6,8 @@ class MCollective::Application::Find<MCollective::Application
 
     starttime = Time.now
 
+    mc.detect_and_set_stdin_discovery
+
     nodes = mc.discover
 
     discoverytime = Time.now - starttime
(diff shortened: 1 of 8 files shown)


MikaelSmith:
Hmm, I get a Rust backtrace from `puppet query` when I try this. I'll follow up with them. Seems good though.

ripienaar:
@MikaelSmith I got that when trying to use it with a app that doesnt support it - not all things support it, `mco rpc` is a safe way to test.

Thing is lets say you have custom app that say do multiple rpc requests, it couldnt possibly use this without making careful decisions about what to do iwth the discovery data etc, so only `rpc` and `find` supports it out the box.

The rust backtrace is when nothing consumed STDOUT from the rust app - like mco apps that dont support this method

MikaelSmith:
I was using `puppet query "inventory { facts.os.name = 'CentOS' }" | mco puppet status` and seeing the error. Shouldn't that be using rpc?

ripienaar:
yeah, `mco puppet` isnt aware of this plugin, try `mco rpc puppet status`.  It's conceivable that we could make some `mco puppet` stuff aware of it easily and others with a bit more work.  Especially now that I turned the thing to enable it into a simple helper

PR #402, opened 2016-10-25
(MCO-777) Wait for all expected responses to rpc, report surprises

MikaelSmith:
Modifies mco to wait for responses from all discovered hosts, rather
than stopping when the expected number of hosts have responded. This
ensures that, if a node doesn't respond to discovery but responds to the
rpc, we don't lose responses from all discovered nodes.

Also update stats to print all responses that came from nodes that did
not respond to discovery. This simplifies debugging issues where nodes
respond unexpectedly, allowing us to identify the nodes behaving
unexpectedly.

### DIFF
--- a/lib/mcollective/client.rb
+++ b/lib/mcollective/client.rb
@@ -159,10 +159,10 @@ module MCollective
     #
     # It returns a hash of times and timeouts for discovery and total run is taken from the options
     # hash which in turn is generally built using MCollective::Optionparser
-    def req(body, agent=nil, options=false, waitfor=0, &block)
+    def req(body, agent=nil, options=false, waitfor=[], &block)
       if body.is_a?(Message)
         agent = body.agent
-        waitfor = body.discovered_hosts.size || 0
+        waitfor = body.discovered_hosts || []
         @options = body.options
       end
 
@@ -174,7 +174,6 @@ module MCollective
       stat = {:starttime => Time.now.to_f, :discoverytime => 0, :blocktime => 0, :totaltime => 0}
       STDOUT.sync = true
       hosts_responded = 0
-
 
       begin
         if threaded
@@ -238,16 +237,39 @@ module MCollective
     def start_receiver(requestid, waitfor, timeout, &block)
       Log.debug("Starting response receiver with timeout of #{timeout}")
       hosts_responded = 0
+
+      if (waitfor.is_a?(Array))
+        unfinished = Hash.new(0)
+        waitfor.each {|w| unfinished[w] += 1}
+      else
+        unfinished = []
+      end
+
       begin
         Timeout.timeout(timeout) do
-          begin
+          loop do
             resp = receive(requestid)
             yield resp.payload
             hosts_responded += 1
-          end while (waitfor == 0 || hosts_responded < waitfor)
+
+            if (waitfor.is_a?(Array))
+              sender = resp.payload[:senderid]
+              if unfinished[sender] <= 1
+                unfinished.delete(sender)
+              else
+                unfinished[sender] -= 1
+              end
+
+              break if !waitfor.empty? && unfinished.empty?
+            else
+              break unless waitfor == 0 || hosts_responded < waitfor
+            end
+          end
         end
       rescue Timeout::Error => e
-        if (waitfor > hosts_responded)
+        if (waitfor.is_a?(Array) && !unfinished.empty?)
+          Log.warn("Could not receive all responses. Did not receive responses from #{unfinished.keys.join(', ')}")
+        elsif (waitfor > hosts_responded)
           Log.warn("Could not receive all responses. Expected : #{waitfor}. Received : #{hosts_responded}")
         end
       end
--- a/lib/mcollective/rpc/stats.rb
+++ b/lib/mcollective/rpc/stats.rb
@@ -2,9 +2,9 @@ module MCollective
   module RPC
     # Class to wrap all the stats and to keep track of some timings
     class Stats
-      attr_accessor :noresponsefrom, :starttime, :discoverytime, :blocktime, :responses, :totaltime
-      attr_accessor :discovered, :discovered_nodes, :okcount, :failcount, :noresponsefrom, :responsesfrom
-      attr_accessor :requestid, :aggregate_summary, :ddl, :aggregate_failures
+      attr_accessor :noresponsefrom, :unexpectedresponsefrom, :starttime, :discoverytime, :blocktime, :responses
+      attr_accessor :totaltime, :discovered, :discovered_nodes, :okcount, :failcount, :noresponsefrom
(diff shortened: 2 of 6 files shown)


lib/mcollective/client.rb
```diff
@@ -238,16 +237,37 @@ def start_publisher(request, publish_timeout)
     def start_receiver(requestid, waitfor, timeout, &block)
       Log.debug("Starting response receiver with timeout of #{timeout}")
       hosts_responded = 0
+
+      if (waitfor.is_a? Array)
```
MikaelSmith: I've tried to preserve the #req API by falling to the old behavior when a number is provided for `waitfor`. The new behavior is to use an array of expected nodes, and wait for responses from those specific nodes.

Identifying unexpected responses is handled in rpc/stats.

lib/mcollective/rpc/stats.rb
```diff
@@ -248,8 +269,26 @@ def no_response_report
...
 
+        result_text.string
+      end
+
+      # Returns a blob of text indicating what nodes responded but weren't discovered
+      def surprise_response_report
+        result_text = StringIO.new
+
+        if @surpriseresponsefrom.size > 0
+          result_text.puts Util.colorize(:red, "Unexpected response from:")
```
MikaelSmith: I'd appreciate suggestions on how to make this clearer. What we're displaying are nodes that didn't respond to discovery, but did respond to the rpc request.

lib/mcollective/client.rb
```diff
@@ -238,16 +237,37 @@ def start_publisher(request, publish_timeout)
...
+              break if !waitfor.empty? && unfinished.empty?
+            else
+              break unless waitfor == 0 || hosts_responded < waitfor
+            end
+          end
         end
       rescue Timeout::Error => e
-        if (waitfor > hosts_responded)
+        if (waitfor.is_a? Array && !unfinished.empty?)
+          Log.warn("Could not receive all responses. Waiting for responses from #{unfinished.keys.join(', ')}")
```
Sharpie: It seems like we're no longer waiting at this point since a Timeout error has been thrown. If true, then something like "Did not receive responses from ..." might be clearer than "Waiting for responses from".

lib/mcollective/client.rb
```diff
@@ -238,16 +237,37 @@ def start_publisher(request, publish_timeout)
...
             yield resp.payload
             hosts_responded += 1
-          end while (waitfor == 0 || hosts_responded < waitfor)
+
+            if (waitfor.is_a? Array)
+              sender = resp.payload[:senderid]
+              if unfinished[sender] <= 1
+                unfinished.delete(sender)
+              else
+                unfinished[sender] -= 1
```
Sharpie: Would this end up adding a "surprise" node to the unfinished hash? If so, that seems like it will cause the RPC command to wait for the entire timeout since `unfinished.empty?` will be stuck returning `false` even if all the initially expected nodes submit responses.

MikaelSmith:
Still need to add decent testing around the changes.

lib/mcollective/client.rb
```diff
@@ -238,16 +237,39 @@ def start_publisher(request, publish_timeout)
...
+              break if !waitfor.empty? && unfinished.empty?
+            else
+              break unless waitfor == 0 || hosts_responded < waitfor
+            end
+          end
         end
       rescue Timeout::Error => e
-        if (waitfor > hosts_responded)
+        if (waitfor.is_a?(Array) && !unfinished.empty?)
+          Log.warn("Could not receive all responses. Waiting for responses from #{unfinished.keys.join(', ')}")
```
Sharpie: At this point, it seems like we're done waiting as a timeout has been thrown. If so, then "Did not receive responses from" might send a clearer message than "Waiting for responses from"

lib/mcollective/client.rb
MikaelSmith: It doesn't in Ruby 2.2.5. If `sender` isn't in `unfinished`, it returns 0 at line 255 and the delete is a no-op.

lib/mcollective/client.rb
MikaelSmith: Yeah, makes sense.

lib/mcollective/client.rb
Sharpie: Ah. Right. `0` would be less than 1. So, this branch is basically there to allow a node to return more than one reply?

MikaelSmith: Identity is recommended but not required to be unique. It's in the docs somewhere.

spec/unit/mcollective/rpc/client_spec.rb
```diff
@@ -737,7 +737,7 @@ module RPC
           msg.expects(:create_reqid).returns("823a3419a0975c3facbde121f72ab61f")
           msg.expects(:requestid=).with("823a3419a0975c3facbde121f72ab61f").times(10)
 
-          stats = {:noresponsefrom => [], :responses => 0, :blocktime => 0, :totaltime => 0, :discoverytime => 0, :requestid => "823a3419a0975c3facbde121f72ab61f"}
+          stats = {:noresponsefrom => [], :surpriseresponsefrom => [], :responses => 0, :blocktime => 0, :totaltime => 0, :discoverytime => 0, :requestid => "823a3419a0975c3facbde121f72ab61f"}
```
ripienaar: `unexpected` might be better than `surprise`? Surprise implies its a mystery, Unexpected implies we know why and did not expect them :) minor nit pick, no biggie.

lib/mcollective/client.rb
```diff
@@ -238,16 +237,39 @@ def start_publisher(request, publish_timeout)
...
         Timeout.timeout(timeout) do
-          begin
+          loop do
             resp = receive(requestid)
             yield resp.payload
             hosts_responded += 1
-          end while (waitfor == 0 || hosts_responded < waitfor)
+
+            if (waitfor.is_a?(Array))
+              sender = resp.payload[:senderid]
```
ripienaar: Nah it's good, this is set by security/base.rb.

Replies have it https://github.com/puppetlabs/marionette-collective/blob/d9038572c31e2d54699868fee82980f61d46060e/lib/mcollective/security/base.rb#L170
Requests have it: https://github.com/puppetlabs/marionette-collective/blob/d9038572c31e2d54699868fee82980f61d46060e/lib/mcollective/security/base.rb#L181

MikaelSmith:
Updated with tests.

ripienaar:
@MikaelSmith this looks good to me but there's a test fail - and I cant see from travis whats actually going on because it looked like it passed?

MikaelSmith:
Kicked it, green now. Something weird happened in Travis, I don't understand it either.

ripienaar:
@MikaelSmith this looks good then assuming it was manually tested

PR #404, opened 2016-11-03
(MCO-778) use publish_timeout from config

adrienthebo:
The Mcollective client class was previously ignoring the publish_timeout
value when specified in the config file, and by default would wait
forever when unthreaded and would crash when threaded. This commit fixes
this by preferring an explicitly set `publish_timeout` value and then
falling back to the config set value.

### DIFF
--- a/lib/mcollective/client.rb
+++ b/lib/mcollective/client.rb
@@ -170,7 +170,7 @@ module MCollective
       threaded = @options[:threaded]
       timeout = @discoverer.discovery_timeout(@options[:timeout], @options[:filter])
       request = createreq(body, agent, @options[:filter])
-      publish_timeout = @options[:publish_timeout]
+      publish_timeout = @options[:publish_timeout] || @config.publish_timeout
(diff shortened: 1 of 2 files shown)


ripienaar:
👍

lib/mcollective/client.rb
```diff
@@ -170,7 +170,7 @@ def req(body, agent=nil, options=false, waitfor=0, &block)
       threaded = @options[:threaded]
       timeout = @discoverer.discovery_timeout(@options[:timeout], @options[:filter])
       request = createreq(body, agent, @options[:filter])
-      publish_timeout = @options[:publish_timeout]
+      publish_timeout = @options[:publish_timeout] || @config.publish_timeout
```
MikaelSmith: For messages from rpc/client.rb, we should be setting @options with `@publish_timeout = initial_options[:publish_timeout] || Config.instance.publish_timeout`. This specifically impacts ping?

PR #407, opened 2016-12-15
(MCO-777) Fix `mco ping` breakage from prior commit

MikaelSmith:
The prior commit for MCO-777 introduced a bug when running `mco ping`.
It would always report `The ping application failed to run: undefined
method `>' for []:Array`, as it accepts all results and then times out.
Fix by ensuring we don't attempt to compare an array to an integer.

### DIFF
--- a/lib/mcollective/client.rb
+++ b/lib/mcollective/client.rb
@@ -267,8 +267,10 @@ module MCollective
           end
         end
       rescue Timeout::Error => e
-        if (waitfor.is_a?(Array) && !unfinished.empty?)
-          Log.warn("Could not receive all responses. Did not receive responses from #{unfinished.keys.join(', ')}")
+        if waitfor.is_a?(Array)
+          if !unfinished.empty?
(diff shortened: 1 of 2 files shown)


lib/mcollective/client.rb
```diff
@@ -267,8 +267,10 @@ def start_receiver(requestid, waitfor, timeout, &block)
           end
         end
       rescue Timeout::Error => e
-        if (waitfor.is_a?(Array) && !unfinished.empty?)
-          Log.warn("Could not receive all responses. Did not receive responses from #{unfinished.keys.join(', ')}")
+        if waitfor.is_a?(Array)
+          if !unfinished.empty?
```
ericwilliamson: Would a comment here explaining why the this isn't a oneliner save future maintainers time / avoid potential regressions?

MikaelSmith: Sure, I'll add something.

MikaelSmith: Too late. The test should prevent future issues, and I think it's pretty straight-forward.

PR #411, opened 2017-01-13
(packaging) Prepare for 2.10.0 release

MikaelSmith:


MikaelSmith:
I'm making an educated guess on when we'll tag it (Jan 23rd).

PR #413, opened 2017-01-31
(maint) Do not build debs/rpms

melissa:
Since we've integrated marionette collective into the puppet-agent
package, we do not want to build deb or rpm packages. We only want the
gem and tarball to be created when we kick off our build automation.
However, as there are some community members that use this automation,
we should keep the relics in this file for future reference. That is why
they are only commented out, and not fully removed.

### DIFF
--- a/ext/build_defaults.yaml
+++ b/ext/build_defaults.yaml
@@ -1,15 +1,15 @@
 ---
 packaging_url: 'git://github.com/puppetlabs/packaging.git --branch=master'
 packaging_repo: 'packaging'
-default_cow: 'base-squeeze-i386.cow'
-cows: 'base-precise-i386.cow base-trusty-i386.cow'
+#default_cow: 'base-squeeze-i386.cow'
+#cows: 'base-precise-i386.cow base-trusty-i386.cow'
 pbuild_conf: '/etc/pbuilderrc'
 packager: 'puppetlabs'
 gpg_name: 'info@puppetlabs.com'
 gpg_key: '7F438280EF8D349F'
 sign_tar: FALSE
 # a space separated list of mock configs
-final_mocks: 'pl-el-6-i386 pl-el-7-x86_64'
+#final_mocks: 'pl-el-6-i386 pl-el-7-x86_64'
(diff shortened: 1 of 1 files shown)


melissa:
cc/ @mwaggett @underscorgan

underscorgan:
@melissa do we have a feel for whether or not anyone depends on these packages?

melissa:
nope, no feel. I don't think so, we certainly don't internally. And we haven't shipped anything in ages. It's just if folks are using this in their own environments, which is why I wanted to leave those settings there commented out. I don't really think anyone is, but there's always the possibility

underscorgan:
@melissa i've always shipped these I think?

melissa:
wait, really? All the packages?

underscorgan:
I may have been wrong in doing this

melissa:
@MikaelSmith just a heads up with this, it looks like we've previously been shipping mco deb and rpm packages. I assume that's happened rather inconsistently though. I know you've been involved with mco recently, but do you have a sense of if folks are using these packages? It's my inclination to not ship them, since mco is provided in the puppet-agent package. However, I could be wrong about that

MikaelSmith:
Were we shipping them to the old repositories, or to PC1? Either way, I'm not aware of any particular use of them, I think it's safe to stop.

ripienaar:
👍 for just supporting AIO like everything else

melissa:
we were shipping them to the old repositories. It looks like puppet-agent 1.0.0 shipped with mco 2.8.1. Since then, we've shipped mco 2.8.1, 2.8.2, 2.8.3, 2.8.4, 2.8.9, and 2.9.0 to the old repos (data from http://apt.puppetlabs.com/pool/trusty/main/m/mcollective/)

melissa:
Fantastic! The file I've modified can be pared down quite a bit more than this, but this minimal change will prevent us from building mco packages in our pipelines in the first place. We can and will still ship the mcollective-client gem and provide source tarballs at http://downloads.puppetlabs.com/mcollective for all new mco releases shipped in puppet-agent

PR #415, opened 2017-02-06
(MCO-785) Fix regression in --batch option

mateusz-gozdek-sociomantic:
It looks like changes introduced in 928ccac63ae2820bc47679264c12e4adf3ba08c6
broke --batch option, because one of stats hash elements is not
initialized as Array and concat function is called on that object which
leads to crash.

This commit will make sure, that unexpectedresponsefrom key is
initialized as Array.

P.S. I think that some unit test case is missing, as they are passing with `--batch` broken.

### DIFF
--- a/acceptance/tests/mco_puppet_batch_count.rb
+++ b/acceptance/tests/mco_puppet_batch_count.rb
@@ -1,0 +1,12 @@
+test_name "mco puppet count" do
+  step "run puppet count"
+    if mco_master.platform =~ /windows/ then
+      mco_bin = 'cmd.exe /c mco.bat'
+    else
+      mco_bin = 'mco'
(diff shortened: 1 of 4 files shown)


MikaelSmith:
Looks like there's a test that needs to be updated for this change. I can fix that up later today if you don't get to it.

mateusz-gozdek-sociomantic:
Feel free to push any required fixes. I don't have much experience in rspec, to you will probably do it much faster. I can also try to fix it myself tomorrow.

MikaelSmith:
I give up trying to spec test this. Acceptance test should cover it.

mateusz-gozdek-sociomantic:
That was fast, thanks!

PR #420, opened 2017-02-28
(maint) Add mco_master if not specified

MikaelSmith:
If not specified in acceptance tests, add the mco_master property to the
last agent config. Assume it's an agent-only node.

Also fixes an ordering-dependent issue in spec tests.

### DIFF
--- a/acceptance/Rakefile
+++ b/acceptance/Rakefile
@@ -129,7 +129,11 @@ EOS
     else
       targets = DEFAULT_TEST_TARGETS
     end
-    raise(USAGE) unless targets.include? 'mco_master'
+
+    if !targets.include? 'mco_master'
+      targets = targets.gsub( /([^\d]+)$/, 'mco_master.\1')
+    end
+
     cli = BeakerHostGenerator::CLI.new([targets, '--disable-default-role', '--osinfo-version', '1'])
     ENV['BEAKER_HOSTS'] = "tmp/#{targets}-#{SecureRandom.uuid}.yaml"
     FileUtils.mkdir_p('tmp')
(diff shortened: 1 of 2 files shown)


MikaelSmith:
@johnduarte

acceptance/Rakefile
```diff
@@ -129,7 +129,11 @@ EOS
     else
       targets = DEFAULT_TEST_TARGETS
     end
-    raise(USAGE) unless targets.include? 'mco_master'
+
+    if !targets.include? 'mco_master'
+      targets = targets[0..-2]+'mco_master.a'
```
johnduarte: We might want to structure this to be more resilient to a trailing node with additional roles. I believe the parser for beaker-hostgenerator looks for values after a numeric architecture definition. So, we may want to append the user defined roles after `mco_master`.

Serving suggestion:
```
targets = targets.gsub( /([^\d]+)$/, 'mco_master.\1')
```

MikaelSmith: Ok, that looks reasonable.

MikaelSmith: Updated.

johnduarte:
:+1: 

This functions as expected
```
bundle exec rake ci:test:aio  SHA=1.8.2  TEST_TARGET=ubuntu1604-64a
...
              - Test Case Summary for suite 'pre_suite' -
       Total Suite Time: 208.83 seconds
      Average Test Time: 23.20 seconds
              Attempted: 9
                 Passed: 8
                 Failed: 0
                Errored: 0
                Skipped: 1
                Pending: 0
                  Total: 9

      - Specific Test Case Status -
...
              - Test Case Summary for suite 'tests' -
       Total Suite Time: 11.53 seconds
      Average Test Time: 2.88 seconds
              Attempted: 4
                 Passed: 3
                 Failed: 0
                Errored: 0
                Skipped: 1
                Pending: 0
                  Total: 4

```

PR #423, opened 2017-03-01
(MCO-786) Fix intermittent spec failures

MikaelSmith:
When aes_security_spec was loaded before psk_spec, PluginManager would
throw a Runtimeerror that the `security_plugin` was already loaded.
Clear the plugin manager in psk_spec in case aes is loaded first.

### DIFF
--- a/spec/unit/mcollective/security/psk_spec.rb
+++ b/spec/unit/mcollective/security/psk_spec.rb
@@ -3,6 +3,11 @@
 require 'spec_helper'
 require 'mcollective/security/psk'
 
+module MCollective
+  # Clear the PluginManager so that security plugin tests do not conflict
+  PluginManager.clear
+end
+
 module MCollective::Security
   describe Psk do
     before do


joshcooper:
Does this fix https://tickets.puppetlabs.com/browse/MCO-786?

MikaelSmith:
Oh yeah, forgot I'd filed a ticket. I'll update the commit message.

spec/unit/mcollective/security/psk_spec.rb
```diff
@@ -3,6 +3,11 @@
 require 'spec_helper'
 require 'mcollective/security/psk'
 
+module MCollective
+  # Clear the PluginManager so that security plugin tests do not conflict
+  PluginManager.clear
```
joshcooper: Would it make sense to move this logic to `spec_helper` in something like:

```ruby
RSpec.configure do |config|
  config.before :each do |test|
    MCollective::PluginManager.clear
  end
end
```

MikaelSmith: That already exists, see https://github.com/puppetlabs/marionette-collective/blob/master/spec/spec_helper.rb#L23. The problem is specifically requiring the source file at line 4 causes the plugin to be loaded in the PluginManager. So the conflict happens before tests even start running.

MikaelSmith:
Updated commit message.

PR #426, opened 2017-03-12
(MCO-790) Allow M::Client users to access M::Message

ripienaar:
This allow client applications that interact with the base M::Client
access to the M::Message object associated with a reply.

These clients are advanced clients such as mco ping where you'd really
would like to know things like network headers and such.

Typical end users interact with mcollective via M::RPC::Client and it's
behaviour has not changed and those users still cannot get access to
this object

### DIFF
--- a/lib/mcollective/client.rb
+++ b/lib/mcollective/client.rb
@@ -249,7 +249,13 @@ module MCollective
         Timeout.timeout(timeout) do
           loop do
             resp = receive(requestid)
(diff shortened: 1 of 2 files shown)


ripienaar:
I am not sure what the right branch is for this, it's backward compatible so I think releasable on whatever is the next release

MikaelSmith:
I think stable would be fine for this.

PR #427, opened 2017-03-14
(MCO-771) Update docs around ActiveMQ certname

MikaelSmith:
Due to a fix in the stomp gem, the ActiveMQ certname must now match the
hostname used to reach it. Update docs to reflect this.


MikaelSmith:
@joshcooper or @Sharpie mind taking a look at this change? Triggered by the fixes around the STOMP gem.

PR #429, opened 2017-03-30
(packaging) Update version for 2.10.3

branan:


Magisus:
Confirmed that this matches the version specified for release in PA 1.10.0.

MikaelSmith:
Going to check for anything that needs to be in the changelog.

PR #431, opened 2017-04-11
(MCO-799) Consider string and symbol keys equivelant

ripienaar:
This has the end effect that:

    c.get_fact(:fact => "cluster")

is the same as:

    c.get_fact("fact" => "cluster")

in the case where the DDL defined the input :fact for the action
:get_fact

Additionally RPC::Request will now intelligently convert request[:foo]
into fetching "foo" from such a request.  We could perhaps munge the
entire incoming request instead of this approach but this seems simpler
to implement

Long term I want to downgrade MCollective network protocols to be JSON
pure to facilitate non Ruby clients and JSON based REST gateways and
just general removal of user supplied YAML.

This changes makes it so that not everyone have to redevelop their DDLs,
Agents and Clients

The one problem case is where someone legit made a input :foo and also
one "foo" on the same action, if all we have is a string to represent
the input - "foo" - we cannot distinguish between them, such agents will
now warn on the client.  I anticipate this to be extreme edge cases as
we've been really good at recommending symbols in all docs and all our
agents

### DIFF
--- a/lib/mcollective/ddl/agentddl.rb
+++ b/lib/mcollective/ddl/agentddl.rb
@@ -183,7 +183,7 @@ module MCollective
       #
       # This is to assist with moving to a JSON pure world where requests might come
       # in from REST or other languages, those languages and indeed JSON itself does
-      # support symbols.
+      # not support symbols.
       #
       # It ensures a backward compatible mode where for rpcutil both of these requests
       # are equivelant
--- a/lib/mcollective/ddl/agentddl.rb
+++ b/lib/mcollective/ddl/agentddl.rb
@@ -166,11 +166,51 @@ module MCollective
         return unless input
 
         input.keys.each do |key|
-          if !arguments.include?(key) && !input[key][:default].nil? && !input[key][:optional]
+          if key.is_a?(Symbol) && arguments.include?(key.to_s) && !input.include?(key.to_s)
+            compat_arg = key.to_s
+          else
+            compat_arg = key
+          end
+
+          if !arguments.include?(compat_arg) && !input[key][:default].nil? && !input[key][:optional]
             Log.debug("Setting default value for input '%s' to '%s'" % [key, input[key][:default]])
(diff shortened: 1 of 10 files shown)


ripienaar:
I did some additional update here so that the entire `RPC::Request` can function in a world of pure string based inputs - and tested this with a JSON pure client this works all the way through.
 
Might be worth holding off merging this one just yet as I think it wont be a huge deal to make the Client work JSON pure in a backward compat way which will help a bit with what we discussed on Slack @MikaelSmith

ripienaar:
This fails on 1.8.7, any objections if I just remove 1.8.7 from the test matrix?

MikaelSmith:
I won't get to this until late next week at the earliest. @puppetlabs/marionette-collective-maintainers

MikaelSmith:
Confusion between symbols and strings for keys seems to come up frequently. This approach makes sense over converting during serialization/deserialization as it doesn't break people who might be using string keys instead of symbols.

ripienaar:
@MikaelSmith I might take a stab at making the code a tad less horrific, but i fear there's no pretty way to do this, useful to have thoughts though, will try do that this week

lib/mcollective/ddl/agentddl.rb
```diff
@@ -166,13 +166,53 @@ def set_default_input_arguments(action, arguments)
...
-            arguments[key] = input[key][:default]
+            arguments[compat_arg] = input[key][:default]
           end
         end
       end
 
+      # Creates a new set of arguments with string arguments mapped to symbol ones
+      #
+      # This is to assist with moving to a JSON pure world where requests might come
+      # in from REST or other languages, those languages and indeed JSON itself does
```
MikaelSmith: This sentence doesn't read clearly to me. "JSON itself does support symbols"?

ripienaar:
@MikaelSmith I've addressed that comment, I think this is probably OK, its a big ugly but that seems par for the course with what its doing.

PR #432, opened 2017-04-20
(MCO-784) Raise log level of messaging

MikaelSmith:
Brings message traffic to info level to make it easier to debug lost
messages.

### DIFF
--- a/lib/mcollective/client.rb
+++ b/lib/mcollective/client.rb
@@ -66,10 +66,7 @@ module MCollective
     # responses and doesn't execute any passed in code blocks for responses
     def sendreq(msg, agent, filter = {})
       request = createreq(msg, agent, filter)
-
-      Log.debug("Sending request #{request.requestid} to the #{request.agent} agent with ttl #{request.ttl} in collective #{request.collective}")
-
-      request.publish
+      publish(request)
       request.requestid
     end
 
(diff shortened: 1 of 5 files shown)


lib/mcollective/client.rb
```diff
@@ -129,7 +126,7 @@ def receive(requestid = nil)
           raise(MsgDoesNotMatchRequestID, "Message reqid #{reply.requestid} does not match our reqid #{requestid}")
         end
 
-        Log.debug("Received reply to #{reply.requestid} from #{reply.payload[:senderid]}")
+        Log.info("Received reply to #{reply.requestid} from #{reply.payload[:senderid]}")
```
Sharpie: This is the only one I'm having second thoughts about as it seems like it could cause the client logs to roll over much faster than the server logs in a large deployment --- which would make it difficult to track sent vs. received over a long period of time.

Perhaps we should leave this one at DEBUG level?

MikaelSmith: That seems like it might be one of the more useful use-cases. However, the mco command is easier to re-run with debug, so that makes sense to me.

lib/mcollective/runner.rb
```diff
@@ -213,7 +213,7 @@ def receiver_thread
 
     # Deals with messages directed to agents
     def agentmsg(request)
-      Log.debug("Handling message for agent '#{request.agent}' on collective '#{request.collective}'")
+      Log.info("Handling message #{request.requestid} for agent '#{request.agent}' in collective '#{request.collective}' from #{request.payload[:senderid]}: #{request.payload[:body]}")
```
MikaelSmith: Line 196 also has some potential to be useful at debug level. It identifies when messages are ignored by filters.

MikaelSmith:
Updated with clearer logging when messages are dropped.

nicklewis:
:+1: This makes sense to me...

MikaelSmith:
@Sharpie any more feedback?

Sharpie:
Looks good to me 👍

PR #433, opened 2017-05-05
(maint) Add testing around exec via Puppet

MikaelSmith:
PE-14681 and MODULES-3125 identified issues running commands via
Puppet exec from other service agents. These issues led to PUP-6675:
selinux prevents processes from writing to files in /tmp if they
don't have a reason to do so. The Puppet service selinux policy
appears to exempt this, but other services that invoke Puppet - such
as mcollective - aren't configured that way and cause silent exec
failures.

PUP-6675 fixes this issue. We add an acceptance test to ensure that it
doesn't recur in pxp-agent.

### DIFF
--- a/acceptance/tests/mco_puppet_exec.rb
+++ b/acceptance/tests/mco_puppet_exec.rb
@@ -1,0 +1,93 @@
+require 'yaml'
+
+test_name "mco puppet exec" do
+
+  testdir = master.tmpdir('mco_puppet_exec')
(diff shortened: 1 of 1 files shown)


MikaelSmith:
Depends on https://github.com/puppetlabs/puppet/pull/5844 being fixed first.

MikaelSmith:
This should be ready to go.

PR #435, opened 2017-05-17
(MCO-804) Skip cipher monkey patch on ruby 2.4+

MikaelSmith:
ly, we appended "!SSLv2" to the SSLContext
DEFAULT_PARAMS[:ciphers] to ensure that mcollective never uses SSLv2,
either from our http client or when using open-uri. However, ruby 2.4
only defines the `:ciphers` array if using openssl < 1.1.0[1]. As a
result, mcollective running on newer systems would hard fail.

Check existence of array before trying to append to it.

[1] ruby/ruby@c9dc016#diff-8406e11e4a42f9de6badcd0f6a6c4262R33

### DIFF
--- a/lib/mcollective/monkey_patches.rb
+++ b/lib/mcollective/monkey_patches.rb
@@ -134,7 +134,9 @@ class OpenSSL::SSL::SSLContext
     DEFAULT_PARAMS[:options] |= OpenSSL::SSL::OP_DONT_INSERT_EMPTY_FRAGMENTS
   end
 
-  DEFAULT_PARAMS[:ciphers] << ':!SSLv2'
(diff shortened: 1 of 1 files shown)


MikaelSmith:
@joshcooper @Magisus this seems to be the same problem as PUP-7383. Mind taking a look?

Magisus:
👍 

Seems kinda gross that we have to carry these around in both spots, but I don't suppose there's a way around that...

joshcooper:
yeah, it's gross. Nit, looks like a typo/copypasta issue in the commit message: "ly, we appended"

MikaelSmith:
Fixed.

PR #438, opened 2017-06-06
(maint) Add test for PA-1168

MikaelSmith:
Adds a test ensuring mcollective servers will reconnect to ActiveMQ
after a hard connection loss. Simulates lost connections by blocking
ports via the firewall.

Depends on https://github.com/puppetlabs/puppet-agent/pull/1071.

### DIFF
--- a/acceptance/tests/mco_reconnect.rb
+++ b/acceptance/tests/mco_reconnect.rb
@@ -1,0 +1,36 @@
+test_name "mco servers should reconnect if cut-off from broker" do
+  skip_test "Can't manipulate firewall rules easily on Windows" if mco_master.platform =~ /windows/
+
+  # Add firewall rules blocking the source ports for the current incoming connections.
+  # When nodes attempt to reconnect, they'll grab a new outgoing port and should bypass the firewall rules.
+  step "disconnect servers" do
+    hosts.reject {|host| host == mco_master}.each do |host|
+      ip = fact_on(host, 'ipaddress')
+      mco_port = on(mco_master, "netstat -anpt|grep 61613|grep ESTAB|grep -Po '#{ip}:\\K\\d+'").stdout.chomp
+      step "blocking traffic to #{ip}:#{mco_port}" do
+        on(mco_master, "iptables -A INPUT -p tcp --sport #{mco_port} -j DROP")
+        on(mco_master, "iptables -A OUTPUT -p tcp --dport #{mco_port} -j DROP")
+      end
+    end
+
+    on(mco_master, 'iptables -L')
+    teardown do
+      on(mco_master, 'iptables -F')
+    end
+  end
+
+  step "check unreachable ping" do
+    on(mco_master, "mco ping") do
+      assert_equal(0, exit_code)
+      assert_match(/^1\s+replies/, stdout)
+    end
+  end
+
+  # It takes 2 heartbeat failures before stomp will retry. The minimum configurable heartbeat is 30 seconds,
+  # so 30 is the minimum time before mcollective could reconnect.
+  sleep 30
+
+  step "check for reconnection" do
+    retry_on(mco_master, "mco ping | grep '#{hosts.size} replies'", {:max_retries => 10, :retry_interval => 5})
+  end
+end
--- a/acceptance/setup/aio/pre-suite/060_Install-mcollective-daemon.rb
+++ b/acceptance/setup/aio/pre-suite/060_Install-mcollective-daemon.rb
@@ -38,6 +38,8 @@ plugin.ssl_server_public = #{mco_confdir}/server.crt
 plugin.ssl_client_cert_dir = #{mco_confdir}/ssl-clients/
 
 connector = activemq
+plugin.activemq.heartbeat_interval = 30
+plugin.activemq.max_hbrlck_fails = 2
 plugin.activemq.pool.size = 1
 plugin.activemq.pool.1.host = activemq
 plugin.activemq.pool.1.port = 61613


MikaelSmith:
This requires cherry-picking up https://github.com/puppetlabs/marionette-collective/commit/3d80356c196c9091ff127633f70eb20c94a65e0e

MikaelSmith:
@johnduarte any chance you can look at this?

acceptance/tests/mco_reconnect.rb
```diff
@@ -0,0 +1,32 @@
...
+    hosts.reject {|host| host == mco_master}.each do |host|
+      ip = fact_on(host, 'ipaddress')
+      mco_port = on(mco_master, "netstat -anpt|grep 61613|grep ESTAB|grep -Po '#{ip}:\\K\\d+'").stdout.chomp
+      step "blocking traffic to #{ip}:#{mco_port}" do
+        on(mco_master, "iptables -A INPUT -p tcp --sport #{mco_port} -j DROP")
+        on(mco_master, "iptables -A OUTPUT -p tcp --dport #{mco_port} -j DROP")
+      end
+    end
+
+    on(mco_master, 'iptables -L')
```
johnduarte: Is this here for diagnostic purposes? If the previous `iptables` commands were successful I am not sure if a list is needed.

acceptance/tests/mco_reconnect.rb
```diff
@@ -0,0 +1,32 @@
...
+    teardown do
+      on(mco_master, 'iptables -F')
```
johnduarte: Question: Should the teardown flush and restart iptables to ensure that the firewall rules are in a known state for the next test? If so, the flush could be an additional intermediary command to prep for the reconnection.

acceptance/tests/mco_reconnect.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  step "check unreachable ping" do
+    on(mco_master, "mco ping") do
+      assert_equal(0, exit_code)
+      assert_match(/^1\s+replies/, stdout)
+    end
+  end
+
+  sleep 30
```
johnduarte: I am unclear what the sleep is for. Shouldn't the `iptables --flush` instantly make amq available again?

acceptance/tests/mco_reconnect.rb
```diff
@@ -0,0 +1,32 @@
...
+      ip = fact_on(host, 'ipaddress')
+      mco_port = on(mco_master, "netstat -anpt|grep 61613|grep ESTAB|grep -Po '#{ip}:\\K\\d+'").stdout.chomp
+      step "blocking traffic to #{ip}:#{mco_port}" do
+        on(mco_master, "iptables -A INPUT -p tcp --sport #{mco_port} -j DROP")
+        on(mco_master, "iptables -A OUTPUT -p tcp --dport #{mco_port} -j DROP")
+      end
+    end
+
+    on(mco_master, 'iptables -L')
+    teardown do
```
johnduarte: IIRC teardowns are queued up and not executed until the test terminates. So, the flush will not occur until after the attempt to reconnect. Which means the test would not be able to pass.

johnduarte:
Attempting the test fails on RHEL 7. As seen in the output below, I don't believe the `iptables` manipulation is working as intended.

```
      mbmfq137niil2k3.delivery.puppetlabs.net (redhat7-64-2) 15:13:14$ iptables -L
        Chain INPUT (policy ACCEPT)
        target     prot opt source               destination
        DROP       tcp  --  anywhere             anywhere             tcp spt:51352

        Chain FORWARD (policy ACCEPT)
        target     prot opt source               destination

        Chain OUTPUT (policy ACCEPT)
        target     prot opt source               destination
        DROP       tcp  --  anywhere             anywhere             tcp dpt:51352

      mbmfq137niil2k3.delivery.puppetlabs.net (redhat7-64-2) executed in 0.04 seconds

    * check unreachable ping

      mbmfq137niil2k3.delivery.puppetlabs.net (redhat7-64-2) 15:13:14$ mco ping
        /opt/puppetlabs/puppet/lib/ruby/gems/2.4.0/gems/stomp-1.3.3/lib/connection/netio.rb:122: warning: constant ::TRUE is deprecated
        /opt/puppetlabs/puppet/lib/ruby/gems/2.4.0/gems/stomp-1.3.3/lib/connection/utils.rb:237: warning: constant ::TRUE is deprecated
        mbmfq137niil2k3                          time=90.05 ms


        ---- ping statistics ----
        1 replies max: 90.05 min: 90.05 avg: 90.05

      mbmfq137niil2k3.delivery.puppetlabs.net (redhat7-64-2) executed in 5.52 seconds

    * check for reconnection

      mbmfq137niil2k3.delivery.puppetlabs.net (redhat7-64-2) 15:13:49$ mco ping | grep '2 replies'
        Trying command 10 times.
      .      .      .      .      .      .      .      .      .      .      .      .        Command `mco ping | grep '2 replies'` failed.
      RuntimeError: Command `mco ping | grep '2 replies'` failed.
...
      Begin teardown

      mbmfq137niil2k3.delivery.puppetlabs.net (redhat7-64-2) 15:15:49$ iptables -F

```

acceptance/tests/mco_reconnect.rb
MikaelSmith: It's just diagnostic.

acceptance/tests/mco_reconnect.rb
MikaelSmith: The iptables rules as written only prevent specific remote ports the agents are connecting with. When they attempt to reconnect, they'll grab a new outgoing port and should bypass the firewall rules.

acceptance/tests/mco_reconnect.rb
MikaelSmith: Not sure why we'd need to restart them.

acceptance/tests/mco_reconnect.rb
MikaelSmith: It takes 2 heartbeat failures before stomp will retry. The minimum configurable heartbeat is 30 seconds, so 30 is the minimum time before mcollective could reconnect. I can add a comment to that effect.

MikaelSmith:
You need a package with the changes from https://github.com/puppetlabs/puppet-agent/pull/1071 for the test to pass. Otherwise stomp doesn't correctly detect the connection failure and will never try to reconnect.

MikaelSmith:
Done. Thanks for the feedback!

PR #439, opened 2017-06-07
(MCO-806) Update vendored systemu

MikaelSmith:
A problem in the vendored systemu version causes "marshal data too
short" errors on Windows when handling encoded text. Update to
4328dce of https://github.com/ahoward/systemu to pick up fixes to read
binary mode and handle ruby.exe paths with spaces.


MikaelSmith:
Spot testing suggests this should work fine.

PR #440, opened 2017-06-07
(MCO-807) Changes to enable minimal rubocop/securitycop scans and to …

jsane:
…convert YAML.load instance to a safer alternate

Re-enabling rubocop/securitycop scan currently only with checks for YAML.Load (as it is rubocop scans were not even running given the cop names being out of sync). Additional securitycops would be introduced in future.
Also includes changes to lib/mcollective/facts/yaml_facts.rb to convert the YAML.load to instead do parse_file and then strip any tags.


lib/mcollective/facts/yaml_facts.rb
```diff
@@ -14,6 +14,29 @@ def initialize
         super
       end
 
+      # Introducing a new helper method to load facts from a file
+      # using safer alternate of doing a parse file following by 
+      # stripping tags
+      # We cannot use YAML.safe_load since it requires Ruby 2.1 and MCO
+      # needs to compatible with older versions.
```
ripienaar: curious why it needs to be?

ripienaar: tests for this would be great too

jsane:
The rspec failures needs to be fixed (even though acceptance tests passed). The change does not look correct.

PR #445, opened 2017-06-27
(MCO-807) Changes to enable minimal rubocop/securitycop scans and to …

jsane:
…convert YAML.load instance to a safer alternate

Re-enabling rubocop/securitycop scan currently only with checks for YAML.Load (as it is rubocop scans were not even running given the cop names being out of sync). Additional securitycops would be introduced in future.
Also includes changes to lib/mcollective/facts/yaml_facts.rb to convert the YAML.load to safe_load where possible.


.travis.yml
```diff
@@ -14,9 +14,3 @@ env:
 
 matrix:
   exclude:
-    - rvm: 2.2.5
```
MikaelSmith: Why run it on every version?

lib/mcollective/facts/yaml_facts.rb
```diff
@@ -23,7 +23,11 @@ def load_facts_from_source
         fact_files.each do |file|
           begin
             if File.exist?(file)
-              facts.merge!(YAML.load(File.read(file)))
+	      if YAML.respond_to? :safe_load
```
MikaelSmith: This seems like a reasonable compromise.

.travis.yml
jsane: Right...since it is source level scan I guess it need not repeat for every ruby version. Will push an update

PR #446, opened 2017-06-28
(MCO-811) Fixnum is deprecated, use Integer

ripienaar:
Ruby 2.4 deprecated Fixnum and Bignum, Integer should now be used

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -9,6 +9,10 @@ else
   gem 'systemu'
 end
 
+if RUBY_VERSION =~ /^2\.4/
+  gem 'json', '~> 2.1.0'
+end
+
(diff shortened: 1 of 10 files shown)


Gemfile
```diff
@@ -9,6 +9,10 @@ else
   gem 'systemu'
 end
 
+if RUBY_VERSION =~ /^2\.4/
+  gem 'json', '~> 2.1.0'
```
MikaelSmith: Why is this needed?

ripienaar: Old gem also not compatible with new ruby. It's native C code references the C version of Fixnum which doesn't compile anymore

PR #447, opened 2017-06-28
(MCO-812) Thread.exclusive is deprecated, use a Mutex

ripienaar:


ripienaar:
Merge this after https://github.com/puppetlabs/marionette-collective/pull/446 and make sure tests are rerun after that merge so 2.4 is tested

PR #454, opened 2017-08-28
(MCO-816) Allow upgrading puppet-agent with mcollective

MikaelSmith:
Since mcollective is part of puppet-agent packaging, upgrading that
package over an mcollective plugin is tricky. Previously the mcollective
systemd service was currently configured in a way that broke that,
because it killed the entire process tree when stopping as part of the
upgrade.

Set `KillMode=process` for the systemd service (similar to how Puppet
and PXP Agent are configured) so that it can be restarted during the
upgrade without killing the upgrade itself.


nicklewis:
If I understand correctly, this addresses the problem by not killing the package manager process that was forked to do the package upgrade. Is that right?

MikaelSmith:
Yes. Without this the entire process tree is killed if mcollective is restarted (which happens during a puppet-agent upgrade). That's the default behavior (`control-group`) for [KillMode](https://www.freedesktop.org/software/systemd/man/systemd.kill.html#KillMode=).

This configuration change ensures only the mcollective process is killed, which means subprocesses started by the agent (such as the actual upgrade) will continue.

For upgrades via puppet, the puppet-agent agent plugin will fork/exec Puppet; that process will continue running after the mcollective service is stopped.

PR #455, opened 2017-09-16
(MCO-817) Allow unquoted function arguments

MikaelSmith:
Allows function arguments to be unquoted, as in `fact(os.family)`.

### DIFF
--- a/lib/mcollective/matcher/scanner.rb
+++ b/lib/mcollective/matcher/scanner.rb
@@ -188,8 +188,9 @@ module MCollective
           return "bad_token", [@token_index - current_token_value.size + 1, @token_index]
         else
           if func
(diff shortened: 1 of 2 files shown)


nicklewis:
This code is just *identifying* function calls. Does the code that later parses the call also properly handle the lack of quotes?

MikaelSmith:
I don't plan to put much effort into this. I think, based on the regex, that this change is purely additive. It seems to work for some simple things, like `fact(os.family)`. It might break in other ways, but I think the risk there is low and can be fixed if it comes up.

MikaelSmith:
It probably would make sense to have a more thorough test. And I can see spaces messing things up. I'll spend a little time on that.

PR #464, opened 2017-12-05
(maint) Ensure projects can handle fedora without the f prefix

mwaggett:
This commit removes the 'fedora-prefix' from the filename in acceptance tests.

### DIFF
--- a/acceptance/lib/puppet/acceptance/install_utils.rb
+++ b/acceptance/lib/puppet/acceptance/install_utils.rb
@@ -126,15 +126,13 @@ module Puppet
         case platform
         when /^(fedora|el|centos)-(\d+)-(.+)$/
           variant = (($1 == 'centos') ? 'el' : $1)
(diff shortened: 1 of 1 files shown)


MikaelSmith:
This is probably fine. I'm not sure we even run tests with mco on fedora.

MikaelSmith:
On that note, not sure it's needed either.

mwaggett:
@MikaelSmith fair, but consistency++ 😊  
I'm gonna throw up a Do Not Merge though, since this should probably go in at the same time as https://github.com/puppetlabs/puppet-agent/pull/1280, https://github.com/puppetlabs/puppet/pull/6414, and https://github.com/puppetlabs/puppetlabs-puppet_agent/pull/232.

mwaggett:
This won't be needed once [MCO-831](https://tickets.puppetlabs.com/browse/MCO-831) goes in.

PR #465, opened 2018-02-08
(maint) Update acceptance tests that reference nightlies

mwaggett:
This commit changes the name of the nightlies server to the actual hostname so that when we remove the 'nightlies' cname (to use for *new* nightlies), tests won't break. THIS SHOULD NOT BE PERMANENT (see RE-10231).


MikaelSmith:
I don't think this runs automatically anywhere, so I'm fine just ignoring it while things change.

mwaggett:
@branan ^ can you confirm? I know you all added all kinds of automation around puppet-agent and its components and I don't wanna break your shit.

PR #467, opened 2018-02-20
(docs) Prepare for 2.12.0 release

MikaelSmith:


website/releasenotes.md
```diff
@@ -8,7 +8,21 @@ This is a list of release notes for various releases, you should review these
 before upgrading as any potential problems and backward incompatible changes
 will be highlighted here.
 
-<a name="2_11.4">&nbsp;</a>
+<a name="2_12_0">&nbsp;</a>
+
+## 2.12.0 - TBD
```
branan: Should we preemptively fill in the puppet-agent 5.5.0 release date here?

MikaelSmith: We could. Going to need to figure out that pattern if we want to keep having release notes live in the code.

branan: I'd honestly rather we didn't, as it puts weird manual bits in all our automatic release processes :)

As for the date, it's up to you. I'm worried we'll forget to do it over the next month and ship with TBD

MikaelSmith: We could stop including the date. Or we can set it here, and not care if it turns out to be wrong.

MikaelSmith: Do we have a release date set?

branan: March 20th is the plan

MikaelSmith: Updated.

PR #471, opened 2018-03-09
(MCO-832) Include offset from UTC in log messages

Sharpie:
This commit updates the format used by the File_logger to include the offset
from UTC in the timestamp added to the log messages. This context helps with
troubleshooting issues that arise when the client, server, and broker are
operating in different time zones.

### DIFF
--- a/lib/mcollective.rb
+++ b/lib/mcollective.rb
@@ -59,7 +59,7 @@ module MCollective
 
   MCollective::Vendor.load_vendored
 
-  VERSION="2.12.0"
+  VERSION="2.13.0"
 
   def self.version
     VERSION
--- a/lib/mcollective/logger/file_logger.rb
+++ b/lib/mcollective/logger/file_logger.rb
@@ -15,6 +15,8 @@ module MCollective
 
         @logger = ::Logger.new(config.logfile, config.keeplogs, config.max_log_size)
         @logger.formatter = ::Logger::Formatter.new
+        # ISO-8601 with sub-second precision and offset from UTC.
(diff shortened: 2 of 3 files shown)


MikaelSmith:
There's a mock logger that needs to be updated.

Sharpie:
Ah yes, got it updated!

lib/mcollective/logger/file_logger.rb
```diff
@@ -15,6 +15,8 @@ def start
 
         @logger = ::Logger.new(config.logfile, config.keeplogs, config.max_log_size)
         @logger.formatter = ::Logger::Formatter.new
+        # ISO-8601 with sub-second precision and offset from UTC.
+        @logger.formatter.datetime_format = "%Y-%m-%dT%H:%M:%S.%6N%:z "
```
MikaelSmith: Is the trailing space intentional?

MikaelSmith: Yup, necessary to leave space for the PID

Sharpie: Yeah, should have called that out in the comment.

lib/mcollective.rb
```diff
@@ -59,7 +59,7 @@ module MCollective
 
   MCollective::Vendor.load_vendored
 
-  VERSION="2.12.0"
+  VERSION="2.13.0"
```
Sharpie: Looks like this might have gone into the 2.12.x branch as part of the re-target.

MikaelSmith: Ugh oops. I'll revert that.

MikaelSmith: Fixed up.

PR #474, opened 2018-03-21
(MCO-833) Fix loading facts when YAML file contains aliases

camilo-schoeningh-sociomantic:
If YAML.safe_load is in use, we need to explicity enable aliases. This
is needed for example by choria-mcollective which is creating
aliases for `libdir`.


mateusz-gozdek-sociomantic:
We also found, that some facts requires `Symbol` class to be serialized, so I think `Symbol` should be whitelisted (added to first array).

MikaelSmith:
Whitelisting symbols would be acceptable.

camilo-schoeningh-sociomantic:
@MikaelSmith Done.

PR #475, opened 2018-03-26
(MCO-817) Only allow simple expressions to be unquoted

MikaelSmith:
A more restrictive approach to MCO-817 to simplify things like structured fact lookup and fstat calls.


MikaelSmith:
This stuff is a little unclear. It seems like there was some attempt to sanitize input in the past because it would be evaled to invoke [data plugins](https://puppet.com/docs/mcollective/current/reference/plugins/data.html). However that was better addressed by MCO-765.

PR #476, opened 2018-04-11
(MCO-834, MCO-836) Document selecting on structured facts, fix Regexp

MikaelSmith:


lib/mcollective/security/ssl.rb
```diff
@@ -224,7 +224,7 @@ def deserialize(msg)
         case serializer
         when "yaml"
           if YAML.respond_to? :safe_load
-            return YAML.safe_load(msg, [Symbol])
+            return YAML.safe_load(msg, [Symbol, Regexp])
```
MikaelSmith: This doesn't appear to affect Choria, so I think this is fine as a fix for PE users.

MikaelSmith:
I may need to retarget this at older releases.

MikaelSmith:
No one appears to have encountered MCO-836, and Choria isn't affected. I think we can just fix in 2.12.x.

PR #479, opened 2018-08-02
(packaging) Use packaging as a gem

mwaggett:
\+ some other small things, see individual commit messages for details


mwaggett:
Not really sure what the deal is with those test failures.. My *guess* is that it's related to the new rake version? but I can't reproduce locally and I don't see "last_comment" anywhere in this repo so idk where that's coming from..

mwaggett:
but this has not come up in other repos where we unpinned the rake version (which is needed for compatibility with packaging)..

MikaelSmith:
Probably mismatch between rake and rspec versions.

mwaggett:
hmm unpinning rspec seems to have gotten rid of the original issue, but now it's saying it can't find mcollective for testing??

Rakefile
```diff
@@ -40,52 +24,19 @@ task :clean do
...
-  Rake::Task["package:deb"].invoke
-
-  if ENV['SIGNED'] == '1'
-    deb_flag = "-k#{ENV['SIGNWITH']}" if ENV['SIGNWITH']
-    safe_system %{/usr/bin/debsign #{deb_flag} pkg/deb/*.changes}
-  end
-  move_artifacts
-end
-
-desc "Build documentation"
```
MikaelSmith: Probably want to leave the doc task.

MikaelSmith:
Can we just skip this PR?? This rspec setup seems to do some terrible things.

mwaggett:
@MikaelSmith we could.. but in the event that we need to do an mcollective release in the future, we would have to ship manually, which we'll no doubt forget when that time comes. I know it's unlikely for us to need another release from this branch (*knock on wood*), but won't later branches have a similar problem? Since we're supporting Puppet Platform 5.5.x in an LTS, which includes mco 2.12.x, I would be surprised if we *never* released again. Do you think this is set up better on later branches? 

and to answer your earlier question, packaging requires rake ~> 12.3

MikaelSmith:
It's not better on later branches.

MikaelSmith:
I think our best bet is to separate rspec and Rake then. Keep rspec pinned to 2.11.z, and change how we run specs to avoid using the Rake task helpers.

mwaggett:
yay! @MikaelSmith the only remaining failures are because packaging doesn't support ruby < 2 anymore.

MikaelSmith:
You can drop those versions from the test matrix.

PR #480, opened 2018-08-06
(maint) Use rspec spec in Travis

MikaelSmith:
To bump to newer versions of rake, we need to stop using it to run rspec
because rspec 2 and rake 12 are incompatible.


MikaelSmith:
Ok, I think this version gives the most readable titles.

